PR #27, opened 2015-01-20
fix wrong exeption message escaping

gschwind:
Hello,

Current escaping of error string escape space as "\ ". This is not necessary, thus I propose to escape using saxutils escaping function that look more apropriate than the base python escape function.

Best regards


jachym:
Thanks

PR #34, opened 2015-04-14
avoid copying files

cehbrecht:
there are three changes in this request:

* using ordereddict for inputs to keep the order of the input parameter as defined.
* if complex type of input parameter is a "file" url try to symlink file (method download2). If not possible download file as before (method download).
* hardlink referenced output to output dir. If not possible copy output.

Linking does not work on windows but in this case it would always fall back to the "copy" behaviour. I could also add a flag so that the "linking" behaviour is deactivated by default.

I need to handle large files (> 1 GB). That is why i would like to avoid unnecessary copies of them.


jachym:
Thank you @cehbrecht

PR #39, opened 2015-10-12
update config form environmental variables

deduikertjes:
I've created some functionality to update config values from environmental variables.

This mainly to be able to set serveraddress and mapserveraddress from the webserver to support virtual hosts. But the mechanism is generic so that any config value can be set from environmental variables.

### DIFF
--- a/pywps/__init__.py
+++ b/pywps/__init__.py
@@ -171,12 +171,12 @@ class Pywps:
     versions=[DEFAULT_VERSION]
     UUID = None
 
-    def __init__(self, method=METHOD_GET, configFiles=None):
+    def __init__(self, environ, configFiles=None):
         """Class constructor
         """
 
         # get settings
-        config.loadConfiguration(configFiles)
+        config.loadConfiguration(configFiles, environ)
         self.setLogFile()
         self.UUID = uuid.uuid1().__str__()
 
@@ -188,7 +188,7 @@ class Pywps:
         DEFAULT_VERSION = self.versions[0]
 
         # find out the request method
-        self.method = method
+        self.method = environ["REQUEST_METHOD"]
 
 
     def parseRequest(self,queryStringObject):
--- a/pywps/config.py
+++ b/pywps/config.py
@@ -57,7 +57,7 @@ def setConfigValue(*args):
 
 
 
-def loadConfiguration(cfgfiles=None):
+def loadConfiguration(cfgfiles=None, environ=None):
     """Load PyWPS configuration from configuration files.
     The later configuration file in the array overwrites configuration
     from the first.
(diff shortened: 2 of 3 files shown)


pywps/__init__.py
```diff
@@ -171,12 +171,12 @@ class Pywps:
     versions=[DEFAULT_VERSION]
     UUID = None
 
-    def __init__(self, method=METHOD_GET, configFiles=None):
+    def __init__(self, environ, configFiles=None):
```
deduikertjes: give an entry for the request environment
Maybe this breaks non-wsgi applications. Maybe something along the lines of:
def __init__(self, environ = {"REQUEST_METHOD":"METHOD_GET"}, configFiles=None):
is better.

pywps/__init__.py
```diff
@@ -171,12 +171,12 @@ class Pywps:
...
         """Class constructor
         """
 
         # get settings
-        config.loadConfiguration(configFiles)
+        config.loadConfiguration(configFiles, environ)
```
deduikertjes: Pass in the request environment as well

pywps/__init__.py
```diff
@@ -188,7 +188,7 @@ def __init__(self, method=METHOD_GET, configFiles=None):
         DEFAULT_VERSION = self.versions[0]
 
         # find out the request method
-        self.method = method
+        self.method = environ["REQUEST_METHOD"]
```
deduikertjes: set the method based on the request environment

pywps/config.py
```diff
@@ -57,11 +57,16 @@ def setConfigValue(*args):
 
 
 
-def loadConfiguration(cfgfiles=None):
+def loadConfiguration(cfgfiles=None, environ=None):
```
deduikertjes: entry point for the request environment to adapt config

pywps/config.py
```diff
@@ -75,7 +80,28 @@ def loadConfiguration(cfgfiles=None):
...
+        _overrideConfigFromConfigDict({k.split("_",2)[-1]:v for k,v in environ.iteritems() if k.startswith('PYWPS_CONFIG_')})
+
+def _overrideConfigFromConfigDict(env_config):
+    # In linux we can have camel case env_vars. In windows we'll run into some trouble.
+    # for now, for windows we'll force section and option to lowercase
+    for k,v in env_config.iteritems():
+        section, option = k.split("_",1)
+        try:
+            if sys.platform == 'win32':
+                config.set(section.lower(), option.lower(), v)
```
deduikertjes: Maybe converting to lower case is not necessary for config sections and options coming from wsgi request environment?

webservices/wsgi/wsgiwps.py
```diff
@@ -54,7 +54,7 @@ def application(environ, start_response):
 
     # create the WPS object
     try:
-        wps = pywps.Pywps(environ["REQUEST_METHOD"])
+        wps = pywps.Pywps(environ)
```
deduikertjes: pass the entire request environment instead of the request method only, so that we can grab settings from there.
this is the way to pass in variables from nginx as nginx is not starting the wsgi app by itself.

PR #41, opened 2015-11-30
Fixed output type handling.

Kruecke:
I tested and found out that the given type checks with `isinstance` are never true (Python 2.7) and so the XML output always defaulted to string. I replaced them with comparisons to `types` so that the results are now as expected.

### DIFF
--- a/pywps/Wps/__init__.py
+++ b/pywps/Wps/__init__.py
@@ -250,18 +250,19 @@ http://wiki.rsg.pml.ac.uk/pywps/Introduction
         :returns: url to w3.org
         """
 
+        import types
         data_type = {"type": None, "reference": None}
-        if isinstance(inoutput.dataType, str):
+        if inoutput.dataType == types.StringType:
             data_type["type"] = "string"
             data_type["reference"] = "http://www.w3.org/TR/xmlschema-2/#string"
-        elif isinstance(inoutput.dataType, float):
+        elif inoutput.dataType == types.FloatType:
             data_type["type"] = "float"
             data_type["reference"] = "http://www.w3.org/TR/xmlschema-2/#float"
-        elif isinstance(inoutput.dataType, int):
+        elif inoutput.dataType == types.IntType:
             data_type["type"] = "integer"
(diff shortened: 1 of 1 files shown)


tomkralidis:
@Kruecke can you provide a use case which demonstrates the problem that this PR fixes?  Note that the [Python docs on using types](https://docs.python.org/2/library/types.html) say to use `isinstance` instead.

cehbrecht:
i suppose `inoutput.dataType` is a `type` not a `str` or `int`. The latter one is checked with `isinstance`.

test:

    import types
    data_type = types.StringType
    data_type == types.StringType
    True
    data_type == types.IntType
    False
    isinstance(data_type, str)
    False
    isinstance(data_type, int)
    False

Kruecke:
@tomkralldis I don't have a lightweight example, but I can tell you what I used this fix for. I have [a boolean flag](https://github.com/bird-house/dodrio/blob/master/processes/wps_cdo.py#L245) in my WPS and I use [Phoenix](https://github.com/bird-house/pyramid-phoenix) to parse an display the WPS as a form to the user. Although Phoenix implements a checkbox as the form element for boolean values, it always displayed textboxes. I figured out that the XML output generated by PyWPS always declared the type to be "string", no matter what I set the type of the flag to. I tracked down the problem and fixed it the way you can see it in this PR.

I hope this is some help to you. Otherwise, a test case that just tries to set the type of any field to anything but string should easily point out this bug. The output will always say "string".

Unfortunately, I'm no python expert so I can't tell you why the previous code didn't work. I can just tell that this fix solved my problem immediately.

tomkralidis:
thanks @Kruecke.  FYI I am unable to reproduce the issue above.  Setting `type` to `type(False)`, or `bool` works correctly.

gkm4d:
Like @cehbrecht stated `inoutput.dataType` is of type `type`, so `isinstance()` does not perform as expected.  I absolutely had to apply this PR to my branch to get `DescribeProcess` to report the correct data type for a Literalnput.

Issue #28, opened 2015-01-27
use of logging is imperfect


doutriaux1:
logging is made via 
```
logging.error("bla")
```

I would suggest adding the capability to use "loggers" rather than logging directly.

I'm trying to use pywps with multiple instance of pywps on the server and it gets confised

I would recommend replacing the logFile in the init with a  logger keyword that would the be used to log, allowing to log simultaneously to multiple logs.

A string with the logger name is actually sufficient you can retrieve the logger via 
```
logging.getLogger(name)
```
Let me know if this makes sense. Thanks for the great work!

cehbrecht:
... when starting with PyWPS i also noticed that pywps logging with self.log() didn't work for async processes. As a workaround i was using my own logging for my processes which also uses concurrent-logging:

https://github.com/bird-house/malleefowl/blob/master/malleefowl/wpslogging.py

But it would be better if we improve logging in PyWPS itself.

jachym:
Hi,

nobody is working on this. We can implement logger in pywps4 and ask
someone to work on this in pywps-3 as for example GSoC project ...

J

Wed Jan 28 2015 at 18:44:04 odesílatel MacPingu <notifications@github.com>
napsal:

tomkralidis:
@doutriaux1: fixed in master.  Feel free to reopen this issue if there are problems.  Thanks!

PR #45, opened 2016-02-05
update logging to use loggers (#28)

tomkralidis:


cehbrecht:
this patched caused failure of logging initialization. `logging.basicConfig()` was replaced by `LOGGER.basicConfig()`. Fixed in patch #49.

tomkralidis:
Thanks @cehbrecht

Issue #70, opened 2016-02-12
Unclosed file handlers lead to OSError when used in wsgi


ricardogsilva:
I am getting

```python
OSError: [Errno 24] Too many open files
```

After leaving my pywps server running for a few days. I am running pywps under wsgi and calling it ~1500 times per day.

After investigating, it seems that everytime that pywps calls `tempfile.mkstemp` an open file handler is left lying around unclosed[1].

According to [python's documentation](https://docs.python.org/2.7/library/tempfile.html#tempfile.mkstemp), `mkstemp` returns a tuple with an already opened file and the path to the file. In pywps we are never using the file handlers (only the file paths) and we never close these open files!

In a wsgi application this becomes a serious problem, because wsgi does some caching of running processes. 

When using apache's mod_wsgi, the process is eventually restarted, causing all file descriptors to be closed. However, if your usage happens to be intensive (as seems to be my use case) there is a chance that you'll run out of file handlers to use.

I'll provide a Pull Request for solving this in the coming days.

[1] - To reproduce the problem, and assuming you get pywps running under wsgi, try the following:

```bash
# reload apache in order to start fresh
sudo service reload apache2

# extract the pid of the wsgi process
ps aux | grep [w]sgi

# check that there are no tmp* files opened by the wsgi process
sudo lsof -p <wsgi_process_pid> | grep tmp

# now run one of your WPS processes

# recheck if there are any tmp* files opened by the wsgi process
sudo lsof -p <wsgi_process_pid> | grep tmp

# there should now be one more tmp* file still open
# everytime you re-run a WPS process another tmp* file will be opened
# until sometime when apache's mod_wsgi decides to clean its cache
```

tomkralidis:
Thanks @ricardogsilva!  FYI your PR is merged in master and backported to branch pywps-3.2.

tomkralidis:
@gkm4d @ricardogsilva fix applied in 7e053dd78bde964557881cb7ea5db74525044af5

PR #71, opened 2016-02-12
fix issue#70, close open file handlers

ricardogsilva:
This pull request fixes issue #70 
All usages of `tempfile.mkstemp()` are now handling the returned open file handler and taking care of closing it afterwards.

Module `pywps.Wps.Execute.UMN` wasn't actually using the generated temporary file for anything, so I removed the creation completely.

### DIFF
--- a/pywps/Process/InAndOutputs.py
+++ b/pywps/Process/InAndOutputs.py
@@ -1014,12 +1014,10 @@ class ComplexOutput(Output):
         elif value.__class__.__name__=='StringIO' or value.__class__.__name__=='StringO':
             import tempfile
             from os import curdir
-            stringIOName = tempfile.mkstemp(prefix="pywpsOutput",dir=curdir) #(5, '/tmp/pywps-instanceS2j6ve/pywpsOutputZxSM6V')
-            stringIOName=stringIOName[1]
-
-            stringIOFile=open(stringIOName,"w")
-            stringIOFile.write(value.getvalue())
-            stringIOFile.close()
+            fh, stringIOName = tempfile.mkstemp(prefix="pywpsOutput",
+                                                dir=curdir) #(5, '/tmp/pywps-instanceS2j6ve/pywpsOutputZxSM6V')
+            fh.write(value.getvalue())
+            fh.close()
             self.value=stringIOName
         # TODO add more types, like Arrays and lists for example
         else:
--- a/pywps/Wps/Execute/__init__.py
+++ b/pywps/Wps/Execute/__init__.py
@@ -1379,11 +1379,14 @@ class Execute(Request):
              # copy the file to safe place
             outName = os.path.basename(output.value)
             outSuffix = os.path.splitext(outName)[1]
-            tmp = tempfile.mkstemp(suffix=outSuffix, prefix="%s-%s" % (output.identifier,self.pid),dir=os.path.join(config.getConfigValue("server","outputPath")))
-            outFile = tmp[1]
-
+            fh, outFile = tempfile.mkstemp(
+                suffix=outSuffix, 
+                prefix="%s-%s" % (output.identifier, self.pid),
+                dir=os.path.join(config.getConfigValue("server", "outputPath"))
+            )
             if not self._samefile(output.value,outFile):
                 COPY(os.path.abspath(output.value), outFile)
+            fh.close()
 
             #check 
             self.contentType = output.format["mimetype"]
--- a/pywps/Wps/Execute/UMN.py
+++ b/pywps/Wps/Execute/UMN.py
@@ -25,7 +25,6 @@ from pywps import config
 import os
 import urllib2
 import logging
-import tempfile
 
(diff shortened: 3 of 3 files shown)


jachym:
Thanks

pywps/Process/InAndOutputs.py
```diff
@@ -1014,12 +1014,10 @@ def setValue(self, value):
...
-            stringIOName=stringIOName[1]
-
-            stringIOFile=open(stringIOName,"w")
-            stringIOFile.write(value.getvalue())
-            stringIOFile.close()
+            fh, stringIOName = tempfile.mkstemp(prefix="pywpsOutput",
+                                                dir=curdir) #(5, '/tmp/pywps-instanceS2j6ve/pywpsOutputZxSM6V')
+            fh.write(value.getvalue())
+            fh.close()
             self.value=stringIOName
```
gkm4d: I concur that the dangling file handles are a legitimate problem, but does this solution actually work?  `fh`, the file handle, is simply an integer.  In my standalone testing with Python 2.6 and 2.7 `fh.write()` and `fh.close()` fail.  Seems like you either need to use: 1)  `fdopen()` to open a file from the descriptor and then write and close the file, or 2) `os.write()` and `os.close()` to operate directly on the file descriptor.

References:
http://stackoverflow.com/questions/7799680/python-need-to-close-file-from-mkstemp-if-i-use-fdopen
http://stackoverflow.com/questions/168559/python-how-do-i-convert-an-os-level-handle-to-an-open-file-to-a-file-object

pywps/Wps/Execute/__init__.py
```diff
@@ -1379,11 +1379,14 @@ def setRawData(self):
...
-            outFile = tmp[1]
-
+            fh, outFile = tempfile.mkstemp(
+                suffix=outSuffix, 
+                prefix="%s-%s" % (output.identifier, self.pid),
+                dir=os.path.join(config.getConfigValue("server", "outputPath"))
+            )
             if not self._samefile(output.value,outFile):
                 COPY(os.path.abspath(output.value), outFile)
+            fh.close()
```
gkm4d: Ditto here.

ricardogsilva:
@gkm4d yes you are absolutely right. I misread the python docs and assumed the return value was a file handler, like the builtin `open` function, while in fact it is an integer (like `os.open`).

Since my use case does not generate any direct outputs, I only tested the correct absence of open file handlers coming from the `pywps.Wps.Execute.UMN` module and missed the error completely.

I can provide a fix for it in the coming days (currently not at the office). Or maybe you'd prefer to PR your fix from:

 https://github.com/gkm4d/pywps/commit/2acbb09aea9b410ea2e2cc3170a75c38b83f8966

Either way, I believe issue #70 should be reopened

tomkralidis:
I've reopened #70.  A fix per above would be valued; we can then issue a bugfix release.

tomkralidis:
@gkm4d @ricardogsilva  does this work (against branch `pywps-3.2`) for everyone?

```diff
diff --git a/pywps/Process/InAndOutputs.py b/pywps/Process/InAndOutputs.py
index 60bbd57..5bdbd14 100644
--- a/pywps/Process/InAndOutputs.py
+++ b/pywps/Process/InAndOutputs.py
@@ -435,7 +435,6 @@ class ComplexInput(Input):
         :param data: the data, which should be stored
         :type data: string
         """
-        import tempfile
         from os import curdir, rename
 
         outputName = tempfile.mktemp(prefix="pywpsInput",dir=curdir)
@@ -1019,12 +1018,11 @@ class ComplexOutput(Output):
         elif type(value) == types.FileType:
             self.value = value.name
         elif value.__class__.__name__=='StringIO' or value.__class__.__name__=='StringO':
-            import tempfile
             from os import curdir
             fh, stringIOName = tempfile.mkstemp(prefix="pywpsOutput",
                                                 dir=curdir) #(5, '/tmp/pywps-instanceS2j6ve/pywpsOutputZxSM6V')
-            fh.write(value.getvalue())
-            fh.close()
+            os.write(fh, value.getvalue())
+            os.close(fh)
             self.value=stringIOName
         # TODO add more types, like Arrays and lists for example
         else:
diff --git a/pywps/Wps/Execute/__init__.py b/pywps/Wps/Execute/__init__.py
index 2d5176e..aa44a15 100644
--- a/pywps/Wps/Execute/__init__.py
+++ b/pywps/Wps/Execute/__init__.py
@@ -1400,7 +1400,7 @@ class Execute(Request):
             )
             if not self._samefile(output.value,outFile):
                 COPY(os.path.abspath(output.value), outFile)
-            fh.close()
+            os.close(fh)
 
             #check 
             self.contentType = output.format["mimetype"]
```

gkm4d:
@tomkralidis Yes, that looks good to me.

jachym:
Nice, thanks @gkm4d

ricardogsilva:
Yes, it seems good now. Thanks a lot!

PR #77, opened 2016-02-15
debian fixes

khosrow:
Fixed debian packaging. Builds for Ubuntu 14.04 and installs ok.


khosrow:
closing. Will submit a new one.

PR #87, opened 2016-02-17
When pickling compiled templates, open file as binary for writing.

gkm4d:
This is consistent with the unpickling operation where the file is
opened as binary, and without this change, under Windows, unpickling
fails.  This change does not seem to have negative implications for
Linux.


jachym:
Should be ok for Linux too, thanks

PR #88, opened 2016-02-17
Base64 encode from source file directly to destination variable

gkm4d:
Use base64.encodestring() to read from file and encode directly to
complexOutput["complexdata"].  Rather than the process of read from file,
base64 encode, write to file, and then read from file again.

Also, use "with" syntax to ensure files are closed after use.


jachym:
Nice, thanks for contribution

jachym:
Fixing https://github.com/geopython/pywps/pull/199
Fixed http://pywps.org/contributors/

Closing

PR #89, opened 2016-02-17
Make minOccurs/maxOccurs checking more robust.

gkm4d:
Fix holes in validating that minOccurs/maxOccurs for inputs are not violated:
1) In _setValueWithOccurence(), check the cases where maxOccurs equals 0 or 1,
in addition to the >1 case.
2) In consolidateInputs() check that minOccurs is always satisfied, including for
the minOccurs>1 case, rather than just checking that there is any single occurrence.

### DIFF
--- a/pywps/Process/InAndOutputs.py
+++ b/pywps/Process/InAndOutputs.py
@@ -118,18 +118,25 @@ class Input:
                 return resp
         return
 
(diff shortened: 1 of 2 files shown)


jachym:
Thanks for contribution

Issue #67, opened 2016-02-09
Adjust PyWPS configuration to pycsw


tomkralidis:
<a href="https://github.com/jachym"><img src="https://avatars.githubusercontent.com/u/1460823?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [jachym](https://github.com/jachym)**
_Monday Jan 25, 2016 at 11:34 GMT_
_Originally opened as https://github.com/PyWPS/pywps-4/issues/83_

----

http://docs.pycsw.org/en/latest/configuration.html

PR #93, opened 2016-02-17
Adjust PyWPS configuration to pycsw (#67)

tomkralidis:
Implements #67


tomkralidis:
@jachym @ldesousa FYI this PR is a breaking change to the codebase / config.  Given the 3->4 path it makes sense.  We should have a migration guide when we start rolling out PyWPS 4.0.0.

Having said this, review/comments/suggestions are valued.  Any objections to merge?

jachym:
Hi, 

pywps-4-demo has to be adjusted as well as documentation (not sure, something about configuration is there, but I believe, it is)

PR #96, opened 2016-02-18
Exception cleanup

gkm4d:
This addresses two issues related to exceptions.

1. The simpler issue is a couple of slight corrections to Execute.tmpl.  First, the `ExceptionReport` element was erroneously in the `wps` namespace when it should be in `ows`.  Secondly, this element was missing a required `version` attribute.  Without these changes, exception outputs would not pass XML schema validation.
2. The second issue is a bit more involved.  Here I have altered `InvalidParameterValue` to take an optional `text` parameter for expressing `ExceptionText`.  This is helpful for indicating the value of the invalid parameter and is recommended in the OGC Web Service Common Standard (see [OGC 06-121r9](http://portal.opengeospatial.org/files/?artifact_id=38867), Page 45, Table 27, Footnote a).  Prior to this change I found it difficult to understand the source of some exceptions that I caused when exercising corner cases.  Furthermore, I have scrubbed much of the code base to add more informative text to `InvalidParameterValue` exceptions where it seemed to make sense, and also ensured that a parameter locator was clearly noted as opposed to a value which was sometimes being used.

### DIFF
--- a/pywps/Exceptions.py
+++ b/pywps/Exceptions.py
@@ -85,6 +85,7 @@ class InvalidParameterValue(WPSException):
     def __init__(self,value,text=None):
         self.code = "InvalidParameterValue"
         self.locator = str(value)
+        self.message = text
         self._make_xml()
         if text:
             self.ExceptionText = self.document.createElement("ExceptionText")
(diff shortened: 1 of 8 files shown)


jachym:
You should join PyWPS-4 development :-D thanks for this - undoubtedly long-term needed fixes.

gkm4d:
@jachym  Thanks.  I've probably got a little too much on my plate right now, but we'll see in the future.  At a minimum I wanted to make sure I shared all the recent tweaks I've made relative to 3.2.

tomkralidis:
+1!

tomkralidis:
+1!

PR #98, opened 2016-02-18
added convenience method getInputValues()

cehbrecht:
the existing getInputValue() returns either a single value or a list. In those cases where i usually expect a list (even with a single value) i use this convenience method getInputValues() so i don't need to check in my processes.


jachym:
@cehbrecht I think you should rebase - or maybe cherry pick your 3 commits on top of current pywps-3.2 branch

PR #106, opened 2016-02-20
close test files

tomkralidis:
# Overview
This PR closes files opened during testing, which emit `ResourceWarning`'s in Python 3.

# Related Issue / Discussion

None

# Additional Information

None

### DIFF
--- a/tests/validator/test_complexvalidators.py
+++ b/tests/validator/test_complexvalidators.py
@@ -64,6 +64,7 @@ class ValidateTest(unittest.TestCase):
         if WITH_GDAL:
             self.assertTrue(validategml(gml_input, MODE.STRICT), 'STRICT validation')
             self.assertTrue(validategml(gml_input, MODE.VERYSTRICT), 'VERYSTRICT validation')
+        gml_input.stream.close()
 
     def test_geojson_validator(self):
         """Test GeoJSON validator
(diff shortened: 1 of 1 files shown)


tomkralidis:
@ricardogsilva @gkm4d: there's a lone `ResourceWarning` in https://travis-ci.org/geopython/pywps/jobs/110529330#L351 that I'm having trouble with. Any suggestions/ideas?

gkm4d:
@tomkralidis I looked at this very quickly.  Here is my suggestion for what to try.  The `test_file()` method of tests/test_inout.py reads:

```
    def test_file(self):
        """Test file input IOHandler"""
        (fd, tmp_file) = tempfile.mkstemp()
        source = tmp_file
        file_handler = open(tmp_file, 'w')
        file_handler.write(self._value)
        file_handler.close()
        self.iohandler.file = source
        self._test_outout(SOURCE_TYPE.FILE)
```

This looks like the same thing recently addressed in Issue #70 where the file is being opened by name and the file descriptor from `mkstemp()` is never closed.  This is a bit removed from the line where the warning is coming from, so perhaps this is not it.  Rewritten code might look like:


```
    def test_file(self):
        """Test file input IOHandler"""
        (fd, tmp_file) = tempfile.mkstemp()
        os.write(fd, self._value)
        os.close(fd)
        self.iohandler.file = tmp_file
        self._test_outout(SOURCE_TYPE.FILE)
```

tomkralidis:
Thanks @gkm4d. Here's some results

- patch above results in
```python
os.write(fd, self._value)
TypeError: 'str' does not support the buffer interface
```
- updating to `os.write(fd, self._value.encode())` results in
```python
.........s.....s........................s/home/tkralidi/work/foss4g/pywps/py3k/pywps/tests/test_inout.py:64: ResourceWarning: unclosed file <_io.FileIO name='/tmp/tmpe15gms4y' mode='rb'>
```

I'll keep digging.

tomkralidis:
Thanks @gkm4d still struggling with this issue.  I'll open a dedicated issue.

Issue #138, opened 2016-06-24
GRASS GIS support


jachym:
# Description

Support for GRASS GIS temporary location and mapset using PyGRASS and GRASS Script interfaces.

* Extend Process class to accept parameter `grass_location`
  - "epsg:1234" - creates new temporary location with given epsg code
  - "/path/to/location/" -  creates new temporary mapset within exiting location (shall be deleted after process calculation is done)
* sets grass environment (makes grass.script and grass.pygrass possible to use) while process handler is called

PR #141, opened 2016-06-26
add support for python and grass gis

jachym:
# Overview

This pull request 

1. Adds support for GRASS GIS in the old (but better) way - the Process can either point to existing location (`grass_location="/path/to/location"`) or create temporary location (`grass_location="epsg:1234"`)

2. Sanitizes inputs to dblog - instead of `insert into ... values {}, {}, {}".format(value1, value2, value3)` `? ? ?` are used in the input of `cursor.execute

3. Fixes cleaning of temporary directory - it is cleaned NOT when the handler finished, but after response is constructed

4. logging LOGGER is now identified by "PYWPS" keyword, rather then `__name__` - all logs should go into one file

# Related Issue / Discussion

#138

# Additional Information

This PR imho contains two critical features - dblog sanitisation and process.workdir cleaning

# Contribution Agreement


jachym:
Work to do:

1. Update documentation
2. Write some tests

PR #147, opened 2016-07-27
web admin and process control

janrvdolf:
Web administration and process control. An untested version of web administration and process control. The REST API will change in near future according to the draft.


cehbrecht:
@jan-rudolf @jachym we probably don't want to merge this PR to master. Should we close it? Should we keep the work as reference somewhere?

jachym:
we did too many changes, before this merge was ready and we could not merge it easily.

I suggest, we close the merge.

I think (and test) possibility to get connected to Django, which gives you many powers for creating gui to pywps

fmigneault:
Closing since very outdated. Feel free to resubmit with updates if still relevant.

PR #191, opened 2016-09-22
added converters for datetime, date and time

cehbrecht:
# Overview

added converter for datetime, date and time to `literaltypes`. Uses `dateutil.parser` to parse datetime strings including time-zone. The previous format string with `%z` did only work on python 3.x, but not on python 2.7.

This patch includes unittests for the datetime converters.

The `python-dateutil` package has been added to the requirements.
# Related Issue / Discussion

see gitter chat.
# Additional Information
# Contribution Agreement

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -9,7 +9,7 @@ Vcs-Git: https://github.com/geopython/pywps.git
 
 Package: python-pywps
 Architecture: all
-Depends: ${misc:Depends}, ${python:Depends}, python-pkg-resources, python-flufl.enum, python-jsonschema, python-lxml, python-owslib, python-werkzeug
+Depends: ${misc:Depends}, ${python:Depends}, python-pkg-resources, python-dateutil, python-flufl.enum, python-jsonschema, python-lxml, python-owslib, python-werkzeug
 Suggests: grass, apache2, apache
 Homepage: http://pywps.org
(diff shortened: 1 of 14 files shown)


tomkralidis:
We should add python-dateutil to Debian deps as well.

cehbrecht:
... i've updated the patch with `python-dateutil` in debian dependencies.

ldesousa:
Can't the Metadata class file be called Metadata.py? What is the logic here?

tomkralidis:
@ldesousa `pywps.app.Common` is in the spirit of OWS Common which PyWPS implements. The idea here is that over time, relevant OWS Common based classes (in addition to Metadata) could be in this class.

ldesousa:
I see. Then, shouldn't this and other OWS common classes be implemented solely in OWSLib?

tomkralidis:
No, OWSLib is for parsing OWS/etc., not serializing.

jachym:
Guys, what is the result from this discussion? It seems to be important patch to me - but I did not follow the discussion completely. Any news/plans?

tomkralidis:
Merged:
- the Metadata class work has already been merged
- the PR introduces a dependency on python-dateutil which @cehbrecht has covered appropriately

Thanks @cehbrecht!

Issue #150, opened 2016-08-05
Error message when a temporary file is removed on Windows 8


geoslegend:
# Description

When a temporary file is removed on Windows 8,  the error message is shown as
```
Traceback (most recent call last):
  File "D:\dev\python27_64\lib\site-packages\pywps-4.0.0a2-py2.7.egg\pywps\app\P
rocess.py", line 297, in clean
    shutil.rmtree(self.workdir)
  File "D:\dev\python27_64\lib\shutil.py", line 252, in rmtree
    onerror(os.remove, fullname, sys.exc_info())
  File "D:\dev\python27_64\lib\shutil.py", line 250, in rmtree
    os.remove(fullname)
WindowsError: [Error 32] : 'd:\\pywps\\temp\\pyws_process_shbh59\\tmpc3n0h0'
```

After 'os.remove(fullname)' syntax was packed by try/catch pattern, I could solve the above problem.
So I would like to request to pack try/catch pattern for users who develop on Windows OS environment.

# Environment

- operating system: Windows 8
- Python version: 2.7.10
- PyWPS version: 4.0.0beta1
- source/distribution
 - [*] git clone
 - [ ] Debian
 - [ ] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [ ] Apache/mod_wsgi
 - [ ] CGI
 - [*] other (please specify): Werkzeug (demo.py)

# Steps to Reproduce

# Additional Information

jachym:
Hi,

how did you get PyWPS downloaded? This should be fixed in current "master" branch as well as in beta1 version

https://github.com/geopython/PyWPS/blob/master/pywps/app/Process.py#L297

geoslegend:
Hi,

I downloaded it on July 22 2016, but today I re-downloaded using the following command and re-tested .

```
git clone https://github.com/geopython/pywps.git
```

However, the same error message is still shown:

```
Traceback (most recent call last):
  File "D:\dev\python27_64\lib\site-packages\pywps-4.0.0a2-py2.7.egg\pywps\app\S
ervice.py", line 368, in _parse_and_execute
    wps_response = process.execute(wps_request, uuid)
  File "D:\dev\python27_64\lib\site-packages\pywps-4.0.0a2-py2.7.egg\pywps\app\P
rocess.py", line 166, in execute
    wps_response = self._execute_process(async, wps_request, wps_response)
[...]

PR #194, opened 2016-10-13
handle file/dir cleanup errors gracefully

tomkralidis:
# Overview
This PR catches cleanup errors, logs the error and emits an `ows:ExceptionReport` response.

# Related Issue / Discussion
#150 
# Additional Information

# Contribution Agreement


ldesousa:
Do we want to the process to fail in case PyWPS fails to remove the temp folder? Wouldn't it be better to send back the response and simply log the exception?

tomkralidis:
Good point @ldesousa.  PR updated to only log the error.

Issue #189, opened 2016-09-20
Support of OWS metadata in process definition


cehbrecht:
In the current pywps-4 implementation one can provide metadata (for process, inputs and outputs) only as a list of strings. These strings become then OWS metadata elements with a `title` attribute.

The OWS metadata element has also the `href` attribute. I extensively use the metadata to provide links to external docs to enhance the process documentation (users need to understand what the process does :)

I provide a patch where one can provide a list of tuples `[(title, href), ]` for metadata in addtion to a list of strings. But i don't know how you would like to have it ...

tomkralidis:
cc @jachym @ldesousa 

@cehbrecht thanks for the report and associated PR in #190.

IMHO it's a good time to change the Process model to make `metadata` a list of dicts, i.e.:

```python
...
metadata=[
    {
        'title': 'foo process',
        'href': 'http://example.org/foo,
    },
    {
        'title': 'foo process more documentation',
        'href': 'http://example.org/bar'
    }
]
...
```

..thus following the OWS metadata model from OWS 1.1.0.

Thoughts?

tomkralidis:
Sample diff:

```diff
diff --git a/pywps/app/Process.py b/pywps/app/Process.py
index e52e746..bfc2ea5 100644
--- a/pywps/app/Process.py
+++ b/pywps/app/Process.py
@@ -112,7 +112,14 @@ class Process(object):
             doc.append(OWS.Abstract(self.abstract))
 
         for m in self.metadata:
-            doc.append(OWS.Metadata({'{http://www.w3.org/1999/xlink}title': m}))
+            metadata = {}
+            if 'href' in m:
+                metadata['{http://www.w3.org/1999/xlink}href'] = m['href']
+            if 'title' in m:
+                metadata['{http://www.w3.org/1999/xlink}title'] = m['title']
+            if 'type' in m:
+                metadata['{http://www.w3.org/1999/xlink}type'] = m['type']
+            doc.append(OWS.Metadata(metadata))
 
         for p in self.profile:
             doc.append(WPS.Profile(p))
```

ldesousa:
According to the specification there can indeed be more than one metadata reference element. But it is defined in another document, I guess that is why no one ever complained about it.

I would prefer to have a Metadata class instead of a list of assorted composition. Does it exist in OWSLib? 

Also, the '{http://www.w3.org/1999/xlink}' string should be promoted to a property to avoid replication.

tomkralidis:
[...]

PR #195, opened 2016-10-14
add support for Process Metadata objects

tomkralidis:
# Overview
This PR provides a proper class for `ows:Metadata` objects, and ensures all metadata class  properties are lists of `pywps.app.Common.Metadata` objects.
# Related Issue / Discussion
#189 
# Additional Information

# Contribution Agreement

### DIFF
--- a/docs/demobuffer.py
+++ b/docs/demobuffer.py
@@ -26,6 +26,7 @@
 __author__ = 'Jachym Cepicky'
 
 from pywps import Process, LiteralInput, ComplexOutput, ComplexInput, Format
+from pywps.app.Common import Metadata
 from pywps.validator.mode import MODE
 from pywps.inout.formats import FORMATS
 
@@ -56,6 +57,7 @@ class DemoBuffer(Process):
             version='1.0.0',
             title='Buffer',
             abstract='This process demonstrates, how to create any process in PyWPS environment',
+            metadata=[Metadata('process metadata 1', 'http://example.org/1'), Metadata('process metadata 2', 'http://example.org/2')])
             inputs=inputs,
             outputs=outputs,
             store_supported=True,
--- a/pywps/app/Common.py
+++ b/pywps/app/Common.py
@@ -1,0 +1,31 @@
+##################################################################
+# Copyright 2016 OSGeo Foundation,                               #
+# represented by PyWPS Project Steering Committee,               #
+# licensed under MIT, Please consult LICENSE.txt for details     #
+##################################################################
+
+
+import logging
+
+LOGGER = logging.getLogger("PYWPS")
+
+
+class Metadata(object):
+    """
+    ows:Metadata content model.
+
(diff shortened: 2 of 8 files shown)


cehbrecht:
@tomkralidis 
i just tried this patch. You probably want to use:

      OWS.Metadata(*dict(m))

instead of 

     OWS.Metadata(dict(m))

tomkralidis:
@cehbrecht thanks for the info. What is the rationale again for this change?

cehbrecht:
ah, the cause for the failure is another ... it does not like the None type which i avoided now by:

     self.href = href or ''

in `pywps.app.Common.Metadata`.

The following code will fail:

       from pywps.app.Common import Metadata
       m = Metadata('fail')
       from pywps import OWS
       OWS.Metadata(dict(m))

This code works:

       from pywps.app.Common import Metadata
       m = Metadata('pass', 'http://somewhere')
       from pywps import OWS
       OWS.Metadata(dict(m))

tomkralidis:
@cehbrecht thanks for the info.  I propose the following instead:

```diff
diff --git a/pywps/app/Common.py b/pywps/app/Common.py
index 4ec2117..a01937e 100644
--- a/pywps/app/Common.py
+++ b/pywps/app/Common.py
@@ -26,5 +26,6 @@ class Metadata(object):
 
     def __iter__(self):
         yield '{http://www.w3.org/1999/xlink}title', self.title
-        yield '{http://www.w3.org/1999/xlink}href', self.href
+        if self.href is not None:
+            yield '{http://www.w3.org/1999/xlink}href', self.href
         yield '{http://www.w3.org/1999/xlink}type', self.type
```

This prevents outputting an empty href attribute in XML. Thoughts?

cehbrecht:
fine :) works for me.

tomkralidis:
@cehbrecht last question: why `OWS.Metadata(*dict(m))` and not `OWS.Metadata(dict(m))`?

cehbrecht:
that was a mistake by myself ... i got these exceptions and i thought it should be `**dict(m)`.

jachym:
@tomkralidis I guess, we should fix pywps-demo, right? 
https://github.com/geopython/pywps-demo/blob/master/processes/sleep.py#L36

tomkralidis:
@jachym done in https://github.com/geopython/pywps-demo/commit/7125c22e3e2a24450e3bc970d9e7045dcc0e70fb

jachym:
Well, it needed a bit more https://github.com/geopython/pywps-demo/commit/b10ff4c0dd6827b7be781e0c30d1244c46896ecd

Side note: IMHO this feature should be discussed more heavily before being committed. We should be in "Feature freeze" stage and the testing was IMHO not sufficient. Just my feeling.

jachym:
He, but thank you for the feature, it's needed!

tomkralidis:
@jachym agree.  Sorry for the oversight.

PR #210, opened 2016-11-06
Adding RELEASE-howto.md file, should fixes #185

### LINKED ISSUE #185: version convention fails Debian packaging
cc @jachym 
# Description

When assembling Debian packaging, the version `4.0.0rc2` is causing error:

```
W: python-pywps: rc-version-greater-than-expected-version 4.0.0rc2 > 4.0.0 (consider using 4.0.0~rc2)
```
# Environment
- operating system: Debian 8
- Python version: 2.7.0
- PyWPS version:
- source/distribution
  - [x] git clone
  - [ ] Debian
  - [ ] PyPI
  - [ ] zip/tar.gz
  - [ ] other (please specify): 
- web server
  - [ ] Apache/mod_wsgi
  - [ ] CGI
  - [ ] other (please specify): 
# Steps to Reproduce

As per https://github.com/geopython/pywps/wiki/ReleasePackaging#debian-packaging
# Additional Information

Changing the version number to `4.0.0-rc2` fixes the issues.

jachym:
# Overview

should fix #185 

# Related Issue / Discussion

# Additional Information

# Contribution Agreement


RELEASE-howto.md
```diff
@@ -0,0 +1,121 @@
...
+
+This document gives you, as PyWPS release master complete tutorial of how to get
+PyWPS release rolled up and deployed to target server, create packages etc.
+
+## PyWPS versioning
+
+PyWPS uses [Debian version naming system](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version). Every policy should be checked against it.
+
+PyWPS uses 3 numbers release description: MAJOR.MINOR.MAINTENANCE. Currently,
+the MAJOR version is `4`. Within MAJOR releases, we should aim, not to break
```
tomkralidis: Suggest we remove the `Currently, the MAJOR version is\`4`` sentence given this will change over time.

RELEASE-howto.md
```diff
@@ -0,0 +1,121 @@
...
+backwards compatibility.
+
+Odd MINOR version numbers (0, 2, 4, 6, ...) are considered as stable, where as
```
tomkralidis: It's actually the reverse, even is stable and odd is unstable/development (see http://producingoss.com/en/development-cycle.html#release-number-even-odd-strategy for more info).

RELEASE-howto.md
```diff
@@ -0,0 +1,121 @@
...
+even numbers (1, 3, 5, 7, ...) are current development branches. MINOR releases
+should add new features.
+
+MAINTENANCE number should be used for bugfix releases only. No new features are
+added.
+
+For release candindates `MAJOR.MINOR.MAINTENANCE~rcX` format should be used.
```
tomkralidis: We should remove any examples with "rc" as this is pretty specific.

RELEASE-howto.md
```diff
@@ -0,0 +1,121 @@
...
+## Create new ticket in repository
+
+At https://github.com/geopython/pywps/issues/new create new ticket with the
+title
+
+```
+PyWPS MAJOR.MINOR.MAJOR~rcX release
+```
+
+At the moment, PyWPS uses `4` as MAJOR version number.
```
tomkralidis: Suggest we remove this sentence given this will change over time.

RELEASE-howto.md
```diff
@@ -0,0 +1,121 @@
...
+
+Create new branch in your private repository
```
tomkralidis: Should this be done within a fork or directly in the repository?

RELEASE-howto.md
```diff
@@ -0,0 +1,121 @@
...
+
+```
+git commit -m"Creating new release of PYWPS X.Y.Z~rcX fixes #TICKET_NUMBER" -a
+git push YOUR_REMOTE
+```
+
+Create pull request.
+
+## After pull request is accepted
+
```
tomkralidis: Suggest to make more clear that at this point we tag in the actual repo, not a forked repo.

RELEASE-howto.md
```diff
@@ -0,0 +1,121 @@
...
+Send PyWPS to http://pypi.python.org repository
+
+```
+cd /tmp
+git clone git@github.com:geopython/pywps.git pywps-4
+cd pywps-4
+git checkout X.Y.Z
+python setup.py bdist_egg upload
+```
```
tomkralidis: We should have a step here that updates the version back to the development version.  As well, we should discuss branching.

For example, once we release 4.0.0:
- we create a pywps-4.0 branch
- we update the VERSION accordingly in master to '4.1-dev'

RELEASE-howto.md
jachym: IMHO .. it's the standard way, how in PyWPS are accepted changes.

jachym:
@tomkralidis updated, will ask PSC members to have a look at it

RELEASE-howto.md
ldesousa: Just replace "Odd" with "Even".

RELEASE-howto.md
```diff
@@ -0,0 +1,121 @@
...
+## PyWPS versioning
+
+PyWPS uses [Debian version naming system](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version). Every policy should be checked against it.
+
+PyWPS uses 3 numbers release description: MAJOR.MINOR.MAINTENANCE. Currently,
+the MAJOR version is `4`. Within MAJOR releases, we should aim, not to break
+backwards compatibility.
+
+Odd MINOR version numbers (0, 2, 4, 6, ...) are considered as stable, where as
+even numbers (1, 3, 5, 7, ...) are current development branches. MINOR releases
```
ldesousa: Replace "even" with "odd".

RELEASE-howto.md
```diff
@@ -0,0 +1,121 @@
...
+PyWPS MAJOR.MINOR.MAJOR-rcX release
+```
+
+At the moment, PyWPS uses `4` as MAJOR version number.
+
+Create new branch in your private repository
+
+```
+git checkout master # or checkout to other point in history, where you need to create new release
+git checkout -b TICKET_NUMBER_pywps-X.Y.Z-rcX-release
```
ldesousa: In the contributing guidelines we make explicit reference to the `master` branch:
http://pywps.readthedocs.io/en/latest/development.html

I would prefer to have a small number of branches and simply a tag for a new release.

RELEASE-howto.md
```diff
@@ -0,0 +1,121 @@
...
+```
+## Fix files, commit and create pull request
+
+* Fix the [VERSION.txt](https://github.com/geopython/pywps/blob/master/VERSION.txt) file.
+* Fix the [pywps/__init__.py](https://github.com/geopython/pywps/blob/master/pywps/__init__.py) file `__version__` attribute
+* Fix the [debian/changelog](https://github.com/geopython/pywps/blob/master/debian/changelog) file
+
+```
+git commit -m"Creating new release of PYWPS X.Y.Z-rcX fixes #TICKET_NUMBER" -a
+git push YOUR_REMOTE
```
ldesousa: I believe this will not push the new branch.

In any case: why do we need this new branch?

RELEASE-howto.md
ldesousa: Tom: what is the purpose of having a new branch for each new release? This can easily become unmanageable.

tomkralidis: @ldesousa we need to issue a new branch with each MINOR release to be able to support a release branch over time.  For example, releasing 4.0.0, or 4.2.0, or 4.4.0 etc. may require bug fixing in a release specific branch which is not applicable / since gone in master branch.

Or the reverse case where a bugfix in master can also be applied to stable branches.  Of course, the project needs some policy on how far back fixes are to be backported.  From my experience, usually the current stable release branch is fine, unless it's a security fix in which case fixes may be backported further.

jachym: I agree - you need place to fix bugs - and merge them back to master too. Not sure how intensively we will do this during our development, but this should be our target procedure.

RELEASE-howto.md
jachym: @ldesousa as already noted: Having branch for MAJOR.MINOR versions should enable us to apply (merge) bug fix to current stable releases - the development tree can move one and we can still be able to support older (still supported) versions  - or "current stable".

RELEASE-howto.md
ldesousa: This increases complexity and the amount of work to keep the repository consistent. What are the gains? If a user needs a bug fix, what is the difference between upgrading to 4.2.1 or 4.4.0, for instance?

tomkralidis: @ldesousa what are the alternatives?  Just keep a master branch and do release tags as needed?

ldesousa: @tomkralidis Yes, I would prefer that simpler approach. All bug fixes would be committed to the `master` branch, developments for a new MAJOR.MINOR.0 release could be committed to a separate branch, say `develop`.

I understand the paradigm you propose, it makes sense. But is it needed in a project of this dimension?

tomkralidis: @ldesousa so something like the [git-flow approach](http://nvie.com/posts/a-successful-git-branching-model/)?  I'm okay with a different development approach which we can agree on here or on the mailing list.

From my experience, as long as we are not forced into a major release bump because we committed a major change to master everytime we want to push out a fix or feature, is or key importance.

ldesousa: @tomkralidis Interesting, I was not aware of that name, but indeed that is how I am used to work on small projects.

The basic idea is to separate feature development from bug fixing (hence the two branches). Therefore `master` can advance as fast or as slow as we wish, which I believe addresses your concerns.

There are two big advantages to this approach:
- No need to worry with backports, i.e. less house keeping work.
- The number of branches is kept at a minimum, i.e. easier to manage.

I propose a PSC vote on this.

tomkralidis: @ldesousa so `master` becomes the moving bug fix branch.  One concern here is what happens when we are a few minor releases down and we come up with an issue of a major bug that needs to be addressed in PyWPS 4.4.x, 4.2.x, etc. Or do we have an expectation that users will always be 'up to date'?

+1 for a PSC discussion/vote, given this is a major development workflow decision.

ldesousa: @tomkralidis Yes, I expect users to address bugs by simply updating to the latest stable. This assumes there are no modifications to the API between each MAJOR.MINOR release. In my experience, this is a fair assumption with PyWPS.

jorgejesus: Just to add my 2cent (also just jumped now to the discussion). I do prefer the option of @ldesousa  that master becomes a moving/growing branch where we commit bug fixes and have major commits for the MAJOR.MINOR branches that are under develop.

jachym:
@ldesousa @tomkralidis  @jorgejesus I have updated the `RELEASE-howto.md` file according to @ldesousa proposal:

no branches for new release, no tickets, just straight-forward development. is that ok ? @jonas-eberle is the last member of PSC who did not review?

jonas-eberle:
@jachym As I have no experience what is a good approach for software versioning based on git, I am fine with the proposal of @ldesousa and your updated `RELEASE-howto.md`.

RELEASE-howto.md
tomkralidis: @ldesousa I'm okay with this in principle.  I do think we should clearly articulate our development workflow/philosophy in documentation.

I'm cc'ing @kalxas for comments / advice on the proposed dev workflow in the context of packaging (given PyWPS is now part of UbuntuGIS/OSGeo Live).

ldesousa: Yes, we must create a development branch and update the guidelines in the docs. I will try to get it ready before the final 4.0.0 release.

Issue #213, opened 2016-12-03
sphinx-build fails for 4.0.0-rc3: PycodeError: tokenizing failed (exception was: TokenError('EOF in multi-line statement', (124, 0)))


sebastic:
# Description
The Debian package build for 4.0.0-rc3 failed due to sphinx-build failing:
```
sphinx-build -W -b html -d _build/doctrees   . _build/html
Running Sphinx v1.4.9
making output directory...
loading pickled environment... not yet created
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 12 source files that are out of date
updating environment: 12 added, 0 changed, 0 removed
reading sources... [  8%] api
reading sources... [ 16%] configuration
reading sources... [ 25%] deployment
reading sources... [ 33%] development
reading sources... [ 41%] exceptions
reading sources... [ 50%] external-tools
reading sources... [ 58%] index
reading sources... [ 66%] install
reading sources... [ 75%] migration
reading sources... [ 83%] process

Exception occurred:
  File "/usr/lib/python2.7/dist-packages/sphinx/pycode/__init__.py", line 259, in tokenize
    raise PycodeError('tokenizing failed', err)
PycodeError: tokenizing failed (exception was: TokenError('EOF in multi-line statement', (124, 0)))
```

# Environment

- operating system: Debian unstable
- Python version: 2.7.11-2
- PyWPS version: 4.0.0-rc3
- source/distribution
 - [ ] git clone
 - [ ] Debian
 - [ ] PyPI
 - [ ] zip/tar.gz
 - [x] other (please specify): GitHub release tarball
- web server
 - [x] Apache/mod_wsgi
 - [ ] CGI
 - [ ] other (please specify): 

# Steps to Reproduce
Build pywps 4.0.0-rc3 in a Debian unstable environment, specifically:
```
cd docs && PYTHONPATH=$(CURDIR) $(MAKE) html
```
Where `$(CURDIR)` is the path to the pywps source toplevel directory

# Additional Information
Attached log file with the full traceback for sphinx-build.
[sphinx-err-mkObpp.txt](https://github.com/geopython/pywps/files/628800/sphinx-err-mkObpp.txt)

sebastic:
I've bisected the commits between rc2 & rc3, and that reveals 98681813c75d0662a3928e0168603e420b331c70 as the commit introducing the issue.

PR #214, opened 2016-12-03
Fix syntax errors causing sphinx-build to fail. (#213)

sebastic:
# Overview
Commit 98681813c75d0662a3928e0168603e420b331c70 introduced syntax errors which caused sphinx-build to fail with:
```
PycodeError: tokenizing failed (exception was: TokenError('EOF in multi-line statement', (124, 0)))
```

# Related Issue / Discussion
#213 

# Contribution Agreement


tomkralidis:
Thanks for the fix @sebastic!

PR #219, opened 2016-12-20
Fix default output url

cehbrecht:
# Overview

the default settings for ``outputurl`` would result in a broken url builded with ``urljoin`` in the storage module. This is fixed with this pull request.

Before:
```
urljoin('file:////tmp/T', 'output.txt') -> file://tmp/output.txt
``` 

Now:
```
urljoin('file:///tmp/T/', 'output.txt') -> file:///tmp/T/output.txt
```

# Contribution Agreement


jachym:
Thanks @cehbrecht, as always

PR #220, opened 2016-12-23
New feature: set HOME to pywps process tempdir

cehbrecht:
# Overview

some tools used in a wps process like to have a ``HOME`` environment and might even write into the ``HOME`` folder (config, cache, ...).

With this new feature the ``HOME`` can be set to the current ``tempdir`` of the process. This feature is disabled by default and can be activated with the ``sethomedir=true`` flag in the pywps server configuration.

You are already using this in ``Process._set_grass``: 
https://github.com/geopython/pywps/blob/master/pywps/app/Process.py#L220

# Contribution Agreement


jachym:
works for me, thanks a lot

PR #222, opened 2017-01-02
Fixing #218 WindowsError on non-windows systems

### LINKED ISSUE #218: Error message when a temporary file failed to be removed on Ubuntu, related to #150
# Description

I'have an error that should be a consequence of some modifications made for #150 :
```
  File "/home/afsy/jrs/Documents/geometwps/srvgeomet27/src/pywps-master/pywps/app/Process.py", line 227, in _run_process
    100, STATUS.DONE_STATUS, clean=self.async)
  File "/home/afsy/jrs/Documents/geometwps/srvgeomet27/src/pywps-master/pywps/app/WPSResponse.py", line 70, in update_status
    self.write_response_doc(self.doc, clean)
  File "/home/afsy/jrs/Documents/geometwps/srvgeomet27/src/pywps-master/pywps/app/WPSResponse.py", line 84, in write_response_doc
    self.process.clean()
  File "/home/afsy/jrs/Documents/geometwps/srvgeomet27/src/pywps-master/pywps/app/Process.py", line 281, in clean
    except WindowsError as err:
NameError: global name 'WindowsError' is not defined
```

# Environment

- operating system: Ubuntu 14.04, 16.04
- Python version: 2.7
- PyWPS version: 4, latest
- source/distribution
 - [x] git clone
 - [ ] Debian
 - [ ] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [ ] Apache/mod_wsgi
 - [ ] CGI
 - [x] other (please specify): Flask

# Steps to Reproduce
Lock on the tmp dir
# Additional Information

jachym:
# Overview

Fixes #218 

# Related Issue / Discussion

Not so stable and robust database logging

# Additional Information

Changes a bit behaviour of database session management too

# Contribution Agreement

### DIFF
--- a/pywps/app/Process.py
+++ b/pywps/app/Process.py
@@ -278,10 +278,8 @@ class Process(object):
             if self._grass_mapset and os.path.isdir(self._grass_mapset):
                 LOGGER.info("Removing temporary GRASS GIS mapset: %s" % self._grass_mapset)
                 shutil.rmtree(self._grass_mapset)
-        except WindowsError as err:
-                LOGGER.error('Windows Error: %s', err)
         except Exception as err:
-                LOGGER.error('Unable to remove directory: %s', err)
+            LOGGER.error('Unable to remove directory: %s', err)
 
     def set_workdir(self, workdir):
         """Set working dir for all inputs and outputs
--- a/pywps/exceptions.py
+++ b/pywps/exceptions.py
@@ -34,7 +34,7 @@ class NoApplicableCode(HTTPException):
     Base exception class
     """
 
-    code = 400
+    code = 500
     locator = ""
 
     def __init__(self, description, locator="", code=400):
--- a/pywps/dblog.py
+++ b/pywps/dblog.py
@@ -25,6 +25,7 @@ from sqlalchemy.orm import sessionmaker
 
 LOGGER = logging.getLogger('PYWPS')
 _SESSION_MAKER = None
+_LAST_SESSION = None
 
 
 _tableprefix = configuration.get_config_value('logging', 'prefix')
@@ -86,6 +87,7 @@ def get_running():
         ProcessInstance.percent_done < 100).filter(
             ProcessInstance.percent_done > -1)
 
+    session.close()
     return running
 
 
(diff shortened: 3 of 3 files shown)


pywps/exceptions.py
```diff
@@ -34,7 +34,7 @@ class NoApplicableCode(HTTPException):
     Base exception class
     """
 
-    code = 400
+    code = 500
```
tomkralidis: The default `code` param below is 400, yet the above line is changed to 500.  Do these need to be consistent?

pywps/app/Process.py
```diff
@@ -278,10 +278,8 @@ def clean(self):
             if self._grass_mapset and os.path.isdir(self._grass_mapset):
                 LOGGER.info("Removing temporary GRASS GIS mapset: %s" % self._grass_mapset)
                 shutil.rmtree(self._grass_mapset)
-        except WindowsError as err:
-                LOGGER.error('Windows Error: %s', err)
         except Exception as err:
-                LOGGER.error('Unable to remove directory: %s', err)
+            LOGGER.error('Unable to remove directory: %s', err)
```
tomkralidis: By deleting the `except` logic won't this result in a fatal error at runtime?

pywps/exceptions.py
jachym: 400 -> NotFound
500 -> ServerError 

which on this place is IMHO more appropriate (no applicable exception)

pywps/exceptions.py
```diff
@@ -34,7 +34,7 @@ class NoApplicableCode(HTTPException):
...
     locator = ""
 
     def __init__(self, description, locator="", code=400):
```
tomkralidis: So here should `code=500`?

pywps/app/Process.py
jachym: `WindowsError` should be subclass of `Exception`, therefore the logic should remain the same. WindowsError is available on windows only - that caused some errors on other platforms (see #218)

this is actually core of the patch. other lines are just minor enhancements

pywps/exceptions.py
jachym: hmm makes sense

btw, just refer to https://en.wikipedia.org/wiki/List_of_HTTP_status_codes we should use them more

Issue #242, opened 2017-03-20
Be able to specify suffix for temporary file


elemoine:
The command I need to wrap into a WPS Process requires that the file names it is passed as input arguments have a specific extension (`.pnm`).

Currently there's no way to specify suffixes for temporary files created by `IOHandler.get_file`.  This issues suggests to extend `IOHandler` and its users to make that possible.

For example, this is what I'd like to be able to do:

```py
inputs = [
    ComplexInput('i', 'image',
                 supported_formats=[Format('image/image/x-portable-anymap')],
                 tempfile_suffix='.pnm')
]
```

Does that sound reasonable? Do you have other suggestions?

Thanks.

jachym:
Hi, that is reasonable reature request. Are you able to dig into this by yourself? Would https://github.com/geopython/pywps/blob/develop/pywps/inout/basic.py#L145 be the line, you are looking for? Of course, we would have to propagate the attribute all the way down to the constructor https://github.com/geopython/pywps/blob/develop/pywps/inout/basic.py#L80 but that should be possible - what do you think?

elemoine:
Yep. Thank you for the note. I'll create a PR for this.

cehbrecht:
From my perspective i have processes which do not only rely on the file extension but also on the filename itself. My wish list would be:

* when a ``ComplexInput`` is provided as url the filename part with extension should be kept (``http://somewhere/path/to/myfile.png  -> myfile.png``). If there are name collisions then the name will be extended ... like with tempfile keeping the filename (without extension) as prefix and the file extension (``myfile-aytgh.png``).
* when the ``ComplexInput`` has no url/filename (base64 included in request) then a tempfile needs to be generated ... as @elemoine said, if would be nice to have an appropriate file extension (``fgbui.png``).

We already have the mime-type formats. Could we associate a default extension to these mime-types which can be used for creating the temp filenames ... and/or add the file extension to the ``Format`` as optional parameter (``Format("image/png", extension=".png")``)?

elemoine:
@cehbrecht thanks for your comments. Did you see https://github.com/geopython/pywps/pull/243? In this patch the extension is specified on the `ComplexInput` itself rather than on the `Format` as you are suggesting. I like your idea of having sensible defaults and storing the extension in the `Format` but I'll have to check that the `Format` object is available when creating the temporary file.

cehbrecht:
@elemoine i saw the PR ... i was a bit late bringing in my comments ;) 

[...]

PR #243, opened 2017-03-21
Make it possible to specify suffix for temp files

elemoine:
# Overview

This PR suggests adding a `tempfile_suffix` argument to `ComplexInput`. This is useful when the program wrapped in a WPS Process requires specific file extensions.

# Related Issue / Discussion

See https://github.com/geopython/pywps/issues/242.

# Contribution Agreement

### DIFF
--- a/pywps/inout/basic.py
+++ b/pywps/inout/basic.py
@@ -142,7 +142,11 @@ class IOHandler(object):
             if self._tempfile:
                 return self._tempfile
             else:
-                (opening, stream_file_name) = tempfile.mkstemp(dir=self.workdir)
+                suffix = ''
+                if hasattr(self, 'data_format') and self.data_format.extension:
+                    suffix = self.data_format.extension
+                (opening, stream_file_name) = tempfile.mkstemp(
+                    dir=self.workdir, suffix=suffix)
                 stream_file = open(stream_file_name, 'w')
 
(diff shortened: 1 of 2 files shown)


jachym:
the patch is OK, simple, straight-forward. Maybe, (you knew, this will come), some tests could be included? `tests/test_inout.py:101 test_file()` ? I guess, 2 lines should fix that.

What do you think about some logic introduced to Format? Wouldn't that be more appropriate? I would find it maybe nicer - ComplexIn/Output can have more formats -> more extensions would be needed. The extension is bound to format (and mimeType).

I know, I should've told you earlier, @cehbrecht brought this topic on right time. Still, could you try to rewrite this? Sorry for inconveniences.

elemoine:
Thanks @jachym. Yes, I'll try using the Format extension.

elemoine:
@jachym this looks tricky to me. The issue is that `IOHandler` doesn't know about the data format. The data format is handled by `BasicComplex`. Did you have something in mind for the implementation of that?

elemoine:
@jachym a bit ugly but works: https://github.com/geopython/pywps/compare/develop...elemoine:suffix2.

elemoine:
@jachym I updated my branch, and added a test. I chose to go the simple path, just setting a suffix when there's a `data_format` with an extension. This may not cover all the use-cases, but it is simple and easy to change/extend in the future.

PR #254, opened 2017-03-29
fixed boundingbox lineage

cehbrecht:
# Overview

Fixed the ``execute_xml`` method of a ``BoundingBoxInput``. 

The WPS 1.0.0 spec is very confusing on bbox parameters. For a execute request (and as such also in the data-input lineage) one uses ``wps:BoundingBoxData`` of type ``ows:BoundingBox``.

See WPS 1.0.0 doc Table 46 and wps schema (BoundingBoxData element):

http://geoprocessing.info/schemas/wps/1.0/wpsExecute_request.xsd

# Contribution Agreement

### DIFF
--- a/pywps/inout/inputs.py
+++ b/pywps/inout/inputs.py
@@ -76,6 +76,8 @@ class BoundingBoxInput(basic.BBoxInput):
         """
         :return: execute response element
         """
+        node = self._execute_xml_data()
+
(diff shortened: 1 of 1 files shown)


cehbrecht:
updated PR ... ``wps:Data`` was missing.

PR #255, opened 2017-03-30
making logging format configurable

cehbrecht:
# Overview

The default logging format is very detailed. This makes it hard to read the logs when debugging user processes. With this patch it is possible to overwrite the default logging format, example:

```
[logging]
format=%(message)s
```

The ``raw=True`` parameter will be used to avoid interpolation by the ``configparser`` for the format option.

# Contribution Agreement

### DIFF
--- a/default-sample.cfg
+++ b/default-sample.cfg
@@ -36,6 +36,7 @@ parallelprocesses=2
 level=INFO
 file=logs/pywps.log
 database=sqlite:///logs/pywps-logs.sqlite3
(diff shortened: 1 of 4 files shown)


tomkralidis:
@cehbrecht thanks for the contribution.  It would be great to have the [docs updated accordingly](https://github.com/geopython/pywps/blob/develop/docs/configuration.rst#logging) as well as part of this PR.

cehbrecht:
@tomkralidis i have added the format option to the docs and the sample config.

PR #260, opened 2017-04-03
using OrderedDict for processes

cehbrecht:
# Overview

Using ``collections.OrderedDict`` (Python > 2.7) the order of the processes will be kept in the GetCapabilities document.

# Additional Information

... was also fixed in pywps 3.x

# Contribution Agreement

### DIFF
--- a/pywps/app/Service.py
+++ b/pywps/app/Service.py
@@ -21,6 +21,7 @@ from pywps.inout.inputs import ComplexInput, LiteralInput, BoundingBoxInput
 from pywps.dblog import log_request, update_response
 
 from collections import deque
+from collections import OrderedDict
 import os
 import sys
 import uuid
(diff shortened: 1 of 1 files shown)


pywps/app/Service.py
```diff
@@ -21,6 +21,7 @@
 from pywps.dblog import log_request, update_response
 
 from collections import deque
+from collections import OrderedDict
```
elemoine: Minor: you can use `from collections import deque, OrderedDict`.

pywps/app/Service.py
```diff
@@ -41,7 +42,8 @@ class Service(object):
     """
 
     def __init__(self, processes=[], cfgfiles=None):
-        self.processes = {p.identifier: p for p in processes}
+        # ordered dict of processes
+        self.processes = OrderedDict([(p.identifier, p) for p in processes])
```
elemoine: Minor: I guess you don't need to create a list. You can just do this instead:

```py
    self.processes = OrderedDict((p.identifier, p) for p in processes)
```

cehbrecht:
@elemoine 
Thanks for review. I've updated the patch according your remarks.

Issue #245, opened 2017-03-27
async requests don't work with sqlite memory database


elemoine:
I've been trying to use async requests (i.e. requests with `storeExecuteResponse`). And I realize that they don't work when an SQLite memory database is used. With an SQLite memory database is used I think the process forked to execute the WPS process gets its own copy of the memory database. So the UPDATE indicating the completion of the process is not seen by the WPS Service process. And because of that the WPS Service will end up queueing requests in the database and eventually rejecting all the requests when `maxprocesses` is reached.

Is this a known issue? What is recommended for async requests? Using an SQLite file database? Something else?

jachym:
@elemoine it's known to me, that sqlite driver is rather unstable and therefore I recommend to use postgres. But your description is giving me some insights of what is really happening.

I personally do not have any strong opinion about this - we need database for logging and running job management, sqlite is file based and does not handle concurency well, I suggest, to use postgres.

Does it answer your question? Do you have some proposal how to overcome this?

elemoine:
I did try Postgres but the SQLAlchemy `BLOB` type used for storing requests doesn't work with Postgres. So at this point my best hope is to use SQLite in file mode.

And no, I have no proposal. But I think it should be documented that PyWPS does not work with SQLite in memory. And I also think that this mode should not be the default. Finally, I'd recommend supporting PostgreSQL.

jachym:
Should be fixed by #268

elemoine:
@jachym have you verified that #268 actually fixes that issue? I have doubts, but I haven't checked myself.

PR #268, opened 2017-05-16
Fix remove stored

cehbrecht:
# Overview

I'm using sqlite database with a datastore file.

This patch fixes the following:
* the ``dblog.remove_stored`` used a wrong argument in ``filter_by``.
* when a process was stored in the queue a status document update was missing.
* capture error when writing lineage of a stored process.
* comments and a try-except block add for running a stored process.  

# Related Issue / Discussion

See also #245 

# Additional Information

Lineage still does not work for stored processes. A stored process is initiated from a json object. This object is used to inititate ``pywps.inout.basic`` input objects but not ``pywps.inout.inputs`` objects which handle the lineage information. See TODO in ``WPSResponse``.

# Contribution Agreement

### DIFF
--- a/pywps/app/Process.py
+++ b/pywps/app/Process.py
@@ -267,16 +267,19 @@ class Process(object):
         # tr
         stored_request = dblog.get_first_stored()
         if stored_request:
-            (uuid, request_json) = (stored_request.uuid, stored_request.request)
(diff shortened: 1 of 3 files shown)


jachym:
Thanks a lot!

PR #269, opened 2017-05-16
New feature: use os.symlink for file:// urls

cehbrecht:
# Overview

In our use-case we are working with climate data netcdf files which can be easily a few GBs large. In a chained process we use file urls as input to a process (data is already available on local filesystem). In the current implementation pywps would **copy** all files to the workdir. We want to avoid that and use **symlinks** instead.

This patch adds a ``file_handler`` to handle complex inputs with file urls and creates symlinks in the workdir to the local file location.

# Related Issue / Discussion

# Additional Information

* This was patched in pywps-3 before.
* ``os.symlink`` works only on unix. One could skip the ``file_handler`` on windows.

# Contribution Agreement

### DIFF
--- a/pywps/app/Service.py
+++ b/pywps/app/Service.py
@@ -396,13 +396,10 @@ class Service(object):
         def href_handler(complexinput, datain):
             """<wps:Reference /> handler"""
             # save the reference input in workdir
-            extension = None
-            if complexinput.data_format:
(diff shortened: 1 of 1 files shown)


jachym:
Cool stuff, thanks a lot @cehbrecht !

PR #274, opened 2017-06-28
New feature: validate complex input file url

cehbrecht:
# Overview

PyWPS allows file URLs as complex inputs. This could be used to access files on the PyWPS server which the processes should not have access to.

With this PR one can configure a list of allowed path locations on the server which are allowed to be used by processes. The path list is configured with the ``allowedinputpaths`` config option, paths can be separated by ``:``. By default no input paths are allowed.

In case of an invalid file URL a ``FileURLNotAllowed`` exception is thrown.

# Related Issue / Discussion

# Additional Information

Maybe this could be implemented as a PyWPS validator ... but it should be enabled for all complex inputs.

# Contribution Agreement

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -119,7 +119,7 @@ configuration file <http://docs.pycsw.org/en/latest/configuration.html>`_.
 :outputurl:
     corresponding URL
 
-.. note:: `outputpath` and `outputurl` must corespond. `outputpath` is the name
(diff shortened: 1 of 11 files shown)


jachym:
I  @cehbrecht,
as @ldesousa is suggesting by your last commit, please consider to merge all 26 commits into one-two commits, so it can be reviewed. This is impractical for further work, unfortunately.

cehbrecht:
I close this pull request in favour of PR #276, which is the same PR but squashed to one commit.

Issue #303, opened 2017-10-12
Bug: binary output does not work with RawDataOutput on Python 3.x


cehbrecht:
# Description

When using the ``RawDataOutput`` parameter in an ``execute`` request for a binary output (image/png) one gets the following exception on Python 3.x:

```
Exception: code: 500, locator: 'utf-8' codec can't decode byte 0x89 in position 0: invalid start byte, description: 
Traceback (most recent call last):
  File "/Users/pingu/sandbox/birdhouse/pywps/pywps/app/Service.py", line 623, in __call__
    request_uuid
  File "/Users/pingu/sandbox/birdhouse/pywps/pywps/app/Service.py", line 309, in execute
    response = self._parse_and_execute(process, wps_request, uuid)
  File "/Users/pingu/sandbox/birdhouse/pywps/pywps/app/Service.py", line 388, in _parse_and_execute
    resp = Response(proc_outpt.data)
  File "/Users/pingu/sandbox/birdhouse/pywps/pywps/inout/basic.py", line 199, in get_data
    content = file_handler.read()
  File "/Users/pingu/.conda/envs/workshop3/lib/python3.6/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x89 in position 0: invalid start byte
```

# Environment

- Python version:
Python 3.x
- PyWPS version:
4.0.0, develop branch
- source/distribution
 - [x] git clone
 
# Steps to Reproduce

Run an execute request with ``RawDataOutput=image_output`` for a binary output.

# Additional Information

See also issue #301

I'm working on a bug-fix.

PR #304, opened 2017-10-13
fix issue #303: RawDataOutput on Python 3.x

cehbrecht:
# Overview

Fixes the issue #303 for RawDataOuput on Python 3.x.

# Related Issue / Discussion

issue #303 

# Additional Information

The patch can optionally use the [python-magic](https://github.com/ahupp/python-magic) library to check if a file is *text* or *binary*.

In PyWPS we already know the mimetype of an output parameter, but I don't know if there is an easy way to identify the mimetype as *text* or *binary*. In this patch all mimetypes with ``text/`` are assumed to be *text*. But mimetype groups can be mixed with both *text* and *binary*,  like ``application/json``  (text) and ``application/x-netcdf`` (binary).

Maybe you have a better idea for dealing with this problem. Python 3.x makes this a bit complicated here.

# Contribution Agreement

### DIFF
--- a/pywps/inout/basic.py
+++ b/pywps/inout/basic.py
@@ -26,6 +26,21 @@ _SOURCE_TYPE = namedtuple('SOURCE_TYPE', 'MEMORY, FILE, STREAM, DATA')
 SOURCE_TYPE = _SOURCE_TYPE(0, 1, 2, 3)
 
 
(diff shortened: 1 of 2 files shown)


jorgejesus:
Fixes #303

PR #307, opened 2018-01-08
Async fixed

lazaa32:
# Overview
Fixed async mode after WPSResponse separation

# Related Issue / Discussion

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/response/execute.py
+++ b/pywps/response/execute.py
@@ -15,7 +15,7 @@ from pywps import WPS, OWS
 from pywps.app.basic import xml_response
 from pywps.exceptions import NoApplicableCode
 import pywps.configuration as config
-import pywps.dblog
+from pywps.dblog import update_response
 
 from pywps.response.status import STATUS
 from pywps.response import WPSResponse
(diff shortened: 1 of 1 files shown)


elemoine:
We're trying to update PyWPS to the latest master in one of our projects. Without this change the "status" document is not generated after an async request. This change does fix the issue. It should probably be merged.

elemoine:
@jachym what do you think about this? Can it be merged?

elemoine:
I am doing more tests, and that actually does not seem to work in all the cases. In the case where the job was stored it looks like the response document is not updated.

pywps/response/execute.py
```diff
@@ -38,6 +38,33 @@ def __init__(self, wps_request, uuid, **kwargs):
...
         self.outputs = {o.identifier: o for o in self.process.outputs}
 
+    def update_status(self, message=None, status_percentage=None, status=None,
+                      clean=True):
+        """
+        Update status report of currently running process instance
+
+        :param str message: Message you need to share with the client
+        :param int status_percentage: Percent done (number betwen <0-100>)
+        :param pywps.app.WPSResponse.STATUS status: process status - user should usually
```
elemoine: The type is not correct. `pywps.response.status.STATUS` should be used instead.

elemoine:
> I am doing more tests, and that actually does not seem to work in all the cases. In the case where the job was stored it looks like the response document is not updated.

Forget this. This was a problem in my own code.

jachym:
Thank you @elemoine for testing. I understand, the code is now better then it was, I'm merging. Please report if any problems should occur in the future.

Issue #297, opened 2017-09-05
How to use format of complex output parameters


cehbrecht:
# Description

The WPS protocol allows you to specify several output formats (mime-types) for a single output parameter. When your process executes you would like to set the output format to a specific one.

Example: allowed formats are PDF and PNG,  in your process you return an output as PDF.

In PyWPS you can set the `output_format` of an output parameter ... but it looks like it has no effect.

Was it intended in that way? Maybe we need to fix it.

# Additional Information

Example process where I set the output format depending on the input:
https://github.com/bird-house/emu/blob/master/emu/processes/wps_inout.py#L142

gschwind:
Hello,

You can set the mime type of an output format with something like:
response.outputs['result'].data_format = pywps.inout.formats.Format(result_mimetype)
response.outputs['result'].set_file("out.csv")

Best regards

cehbrecht:
@gschwind 
you are right, I need to use the `data_format` method instead of `output_format`. This has probably changed at some point. 

But ... doing so leads to exceptions. I have figured them out and fixed things in patch #314 .

tomkralidis:
Fixed in #314

PR #314, opened 2018-01-17
Fix issue #297: output format

cehbrecht:
# Overview

The method to set the output format of an output parameter has changed from `output_format` to `data_format`. But these changes weren't made consistently . This patch fixes the following:

* `output_format` was still used in the `storage` module:
https://github.com/geopython/pywps/blob/3f2e6c63e6cf9d9cd0ad6e690271259205e57ae6/pywps/inout/storage.py#L117

... also in `test_ows.py`.

* In addition the pre-defined `FORMATS` for mime-types were defined as `namedtuple` but within the pywps code it is always assumed that these Formats are `Format` objects (to set and use a validator). This has been fixed by this patch as well.

* It handles also the case when the `extension` attribute of a `Format` object is `None` ... an empty string will be returned then, like for `schema` etc, to avoid exceptions on string concatenation.

* pep8 formatting in the changed files.

# Related Issue / Discussion

#297 

# Additional Information

The `FORMATS` definition works after applying this patch but I suppose we should refactor it later on.

# Contribution Agreement

### DIFF
--- a/docs/demobuffer.py
+++ b/docs/demobuffer.py
@@ -48,6 +48,7 @@ out_output = ComplexOutput(
 inputs = [inpt_vector, inpt_size]
 outputs = [out_output]
 
+
 class DemoBuffer(Process):
     def __init__(self):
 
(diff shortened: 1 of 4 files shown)


idanmiara:
I've just stumbled upon this issue now in v4.5.
Is `output_format` used for anything ? 
It is used in many processes in `pywps-flask` like:
https://github.com/geopython/pywps-flask/blob/master/processes/buffer.py#L77

Maybe we should make it somehow a synonym for `data_format` for backwards compatibility ?

cehbrecht:
@idanmiara we could add `output_format` as synonym for `data_format`:
https://github.com/geopython/pywps/blob/77aa30a2d5e72868099308301a3e1dbfd8fbcedc/pywps/inout/basic.py#L754

But I would prefer to update `pywps-flask`. There are probably more things broken ...

PR #334, opened 2018-03-29
support GRASS location based on georeferenced file

### LINKED ISSUE #333: New Feature: Support GRASS location based on georeferenced file
# Description
In the current version, GRASS location can be defined either by directory name or EPSG. It would be helpful to support also georeferenced file-based location.

pesekon2:
fix #333


ldesousa:
Interesting feature, thank you for contributing. We need the tests though, I will create a new issue and assign it to you, if you do not mind.

PR #341, opened 2018-05-18
Disabled connection pooling to make pywps fork()/clone() safe

ldperron:
# Overview

The following commit solves random errors when using pywps with postgresql+psycopg2 as database backend for the dblog module, along with the multiprocessing processing mode. The default SQLAlchemy creates a connection pool by default, and must not be shared with child processes.

Disabling pooling fixes errors occurring randomly such as:
> ResourceClosedError: This result object does not return rows. It has been closed automatically.

> (psycopg2.DatabaseError) error with status PGRES_TUPLES_OK and no message from the libpq 

# Contribution Agreement

- [X] I'd like to contribute bugfix to PyWPS. I confirm that my contributions to PyWPS will be compatible with the PyWPS license guidelines at the time of contribution.

### DIFF
--- a/pywps/dblog.py
+++ b/pywps/dblog.py
@@ -21,6 +21,7 @@ import sqlalchemy
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy import Column, Integer, String, VARCHAR, Float, DateTime, LargeBinary
 from sqlalchemy.orm import sessionmaker
+from sqlalchemy.pool import NullPool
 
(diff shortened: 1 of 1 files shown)


cehbrecht:
Adding some links for background info:
* http://initd.org/psycopg/docs/usage.html#thread-safety
* http://docs.sqlalchemy.org/en/latest/core/pooling.html#sqlalchemy.pool.NullPool

cehbrecht:
@ldperron merged ... thanks :)

PR #359, opened 2018-07-13
Ensure exception provide the correct Content-Type

gschwind:
Hello,

Because XML is 'text/*' we should set the charset in content type header
this the purpose of this changes.

# Contribution Agreement

### DIFF
--- a/pywps/exceptions.py
+++ b/pywps/exceptions.py
@@ -12,6 +12,7 @@ http://lists.opengeospatial.org/pipermail/wps-dev/2013-October/000335.html
 """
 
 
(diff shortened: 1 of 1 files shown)


gschwind:
To pass unit test this patch series need #361 

Best regards

cehbrecht:
This PR removes `get_headers` method ... but it was not used anywhere.

cehbrecht:
@gschwind: looks simple enough ... so I merge it. Thanks :)

gschwind:
Hello,

The _get_headers_ function was used because it overrides _HTTPException.get_headers_

Best regards

PR #360, opened 2018-07-13
Fixing WPSRequest parsing when no inputs is provided

gschwind:
# Overview

Hello,

When an execute request with GET method has not input an incorrect exception is raised. A WPS process may have no inputs.

This patch fix this issue.

Best regards

# Contribution Agreement


cehbrecht:
@gschwind: thanks again.

PR #361, opened 2018-07-13
Fixing unit test to accept text/xml; charset=*

gschwind:
The unit test is to restrictive and do not accept the valid content type
text/xml; charset=utf-8. This patch correct it

# Contribution Agreement


cehbrecht:
PR looks simple enough ... so I merged it. @gschwind thanks.

Issue #362, opened 2018-07-16
optional metadata href+role is missing in jinja templates


cehbrecht:
# Description

With PR #345 new templates using jinja where introduced. The optional Metadata href and role attributes are not part of these templates. This PR will fix this.

See `Metadata` class: 
https://github.com/geopython/pywps/blob/abdd4a3f01404e98d2d1705663537f9c2d1e5552/pywps/app/Common.py#L11

jachym:
Done, thank you

PR #363, opened 2018-07-16
added metadata href+role (#362)

cehbrecht:
# Overview

This PR is a fix for #362 and adds the optional metadata attributes `href` and `role` to the jinja templates. These attributes are only added when they are set.

Also `autoescape` is enabled for the jinja environment which is needed to encode URLs. 

# Related Issue / Discussion

#362.

# Additional Information

# Contribution Agreement


jachym:
Works for me, thank you @cehbrecht

Issue #370, opened 2018-08-08
workdir is removed before process has started


cehbrecht:
# Description

The process `workdir` is removed *before* the process has started. 

The suspicious line of code is here:
https://github.com/geopython/pywps/blob/fbd5b3a39f2f7b6da735b94bd57caad12c267a8c/pywps/response/execute.py#L209

When I disable it then it works.

@gschwind @jachym Could you please check the reasoning for this code line? Otherwise I would just comment it with a PR.

If your process has `ComplexInputs` then the `workdir` is created again and you might not notice this bug.

# Steps to Reproduce

Try a process with no `ComplexInputs` and which creates a file in `self.workdir`:
https://github.com/cehbrecht/ecas-wps-demo/blob/759a83d390ef28641825eaa1aafde791a4c4a815/ecaswps/processes/wps_index_tn.py#L77

# Additional Information

The mentioned line of code was introduced in PR #358.

gschwind:
Hello, as first thought, I think cleanup should happen as the queue of process is working, i.e. cleaning directories each time services are called, checking for obsolete directories and removing them.

There is also a lot of issue related to the size of the temporary directory.

jachym:
Would something like 

```
209                if self.status == WPS_STATUS.FAILED or self.status == WPS_STATUS.SUCCEEDED:
                        .....

```

help ?

jachym:
Did this apply for synchronous process call as well or only for async mode?

cehbrecht:
There is already a method calling `clean`:
https://github.com/geopython/pywps/blob/2c48403d20cb9c1cc3f7bcc8b0730d468ac763b8/pywps/response/execute.py#L79

Maybe the status updates are not called in the right way.

PR #375, opened 2018-08-14
disabled process clean (#370)

cehbrecht:
# Overview

This PR is a workaround for issue #370. 

It disables the `process.clean()` call to avoid removing the workdir before the process job has started.

It is probably not a solution to #370 and there is more work do be done for process cleaning.

# Related Issue / Discussion

See #370 and #358.

# Additional Information

# Contribution Agreement


cehbrecht:
@gschwind I pull this workaround into master to have a working pywps. Please complain or provide a new PR if you don't like this.

Issue #352, opened 2018-06-29
Add url property to IOHandler


huard:
# Description

In addition to file, memory_object, stream and data, add a url property to IOHandler class to facilitate handling of remote access data protocols.

cehbrecht:
This is related to OpenDAP handling (#152). See also discussion in PR: https://github.com/bird-house/pywps/pull/1

cehbrecht:
The URL property makes sense when we want to support OpenDAP (and other data service protocols). Otherwise we would misuse the `file` property.

See also the discussion in PR https://github.com/bird-house/pywps/pull/2

huard:
I have a question about ComplexOutputs. 

If the output source type is file, and as_reference is True, the output file will be copied over the outputpath (by FileStorage) and the server will send back the url to this file. 
If the output source type is file and as_reference is False, the _json_data tries to parse the file (not sure I understand why it assumes that it is an xml file), and if that fails, it sends back the text or bytes. 

Now what should happen if the source_type is url ? My assumption would be that if as_reference is True, then only the href is sent back, but if as_reference is False, then it would download the href and extract the data. Correct ?

cehbrecht:
> Now what should happen if the source_type is url ? My assumption would be that if as_reference is True, then only the href is sent back, but if as_reference is False, then it would download the href and extract the data. Correct ?

Yes, I would expect the same behavior. I haven't checked but this is already taken care of?

huard:
Yes, in the last commit.

PR #376, opened 2018-08-14
IOHandler refactor

huard:
# Overview

Refactor IOHandler to facilitate adding support for URL. 

The current code in master handles file, data, stream sources using if statements, which can get hairy when adding a new source (e.g. url). This PR creates source-specific handlers. Objects can be created using the basic IOHandler, then specialized at runtime into a FileHandler, DataHandler, StreamHandler and eventually an UrlHandler or a MemoryHandler. 

The main behavioral change is that once a ComplexInput sets a property (file, data, stream), it's not possible for the object to switch to another source type, so for example:

```
h = IOHandler()
h.data = 'test' # h is now an instance of DataHandler
f.file = 'test.txt'  # will raise an error because it would require a switch to a FileHandler class.
print(h.file)        # will work. 
```

The plan is then to add a UrlHandler that would allow an input to set the url property, without downloading the content of the link. What is planned is something like that: 
```
h = IOHandler()
h.url = 'http://...' # Mixes the instance with the UrlHandler class.
h.file # Fetches the data and copies to a local file
h.data # Reads the content from h.file
```

Very few changes were made to app/Service.py, but there is potential to simplify the code with this refactoring. 

# Related Issue / Discussion
Email threads on pywps-dev
https://github.com/geopython/pywps/issues/352

# Additional Information
- This will help support opendap mimetype, which would use the UrlHandler class.  
- No modifications were done to tests, so these changes should be compatible with code out there, except for cases where IOhandler instances set different properties (which is probably a bad idea anyway). 
- I've run the emu test suite with these changes. 

# Contribution Agreement

### DIFF
--- a/pywps/app/Service.py
+++ b/pywps/app/Service.py
@@ -204,7 +204,6 @@ class Service(object):
                 raise NoApplicableCode(e)
 
             complexinput.file = tmp_file
-            assert complexinput.url == datain.get('href')
 
         def file_handler(complexinput, datain):
             """<wps:Reference /> handler.
@@ -228,7 +227,6 @@ class Service(object):
                 shutil.copy2(inpt_file, tmp_file)
 
             complexinput.file = tmp_file
-            assert complexinput.url == datain.get('href')
 
         def data_handler(complexinput, datain):
             """<wps:Data> ... </wps:Data> handler"""
@@ -414,7 +412,7 @@ class Service(object):
         except HTTPException as e:
             return NoApplicableCode(e.description, code=e.code)
         except Exception as e:
-            return NoApplicableCode("No applicable error code, please check error log", code=500)
+            return NoApplicableCode("No applicable error code, please check error log.", code=500)
 
     @Request.application
     def __call__(self, http_request):
--- a/pywps/app/Service.py
+++ b/pywps/app/Service.py
@@ -204,8 +204,7 @@ class Service(object):
                 raise NoApplicableCode(e)
 
             complexinput.file = tmp_file
-            complexinput.url = datain.get('href')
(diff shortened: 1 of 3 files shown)


cehbrecht:
I have double-checked this PR with the Emu processes. Looks all fine :)

I find this PR very useful. It makes sense to have a single IO data type (set via data=, file=, url=) and having several methods to access it (data(), url()). It looks easy to extend and brings us close to our goal to handle OpenDAP. 

I think this PR is safe enough to pull it in for 4.2.

@jachym I leave the final word to you :)

@huard I've activated codacy checks. Please have a look at the link above (a check like travis). Maybe you can make Codacy happier :)

cehbrecht:
@huard tests are now not happy:
`TypeError: _openmode() missing 1 required positional argument: 'data'`

huard:
Hold on. For some reason, relevant tests were skipped. Investigating.

huard:
Fixed in #382

huard:
I found a bug in my refactoring, but fixing it requires a decision from above. 

In the current version, it is not allowed to change source_type for a ComplexOutput once it has been set.  If it's a file, then we can access its data attribute, but not set it. However, we can set the file to another value. The bug is that changing the file attribute does not reset _data and _stream attributes to None, which would mess with the cache and lazy loading. 

The solution is either to allow changing the file property and reset the caches for the other attributes, or blocking developers from modifying the attribute once it's set. That is, you'd have only once chance for a given ComplexInput to set the file/data/stream/memory property, and after that access would be read-only. 

Let me know how you see this so I can update the PR.

huard:
The code change in both case is minor, so this is not an issue. Done.

Issue #381, opened 2018-08-17
Some IOHandler tests are skipped 


huard:
# Description
test_file, test_data, test_stream and test_memory are skipped because of a self.skip statement in _test_outout. 

# Environment
- operating system: Ubuntu 16.04
- Python version: 2.7
- PyWPS version: master
- source/distribution
 - [x] git clone
 - [ ] Debian
 - [ ] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [ ] Apache/mod_wsgi
 - [ ] CGI
 - [ ] other (please specify): 

# Steps to Reproduce
`pytest -rsx tests/test_inout.py`

# Additional Information

PR #382, opened 2018-08-17
removed self.skipTest in _test_outout method.

huard:
# Overview
Removes self.skipTest call in _test_outout. 

# Related Issue / Discussion
#381 

# Additional Information

# Contribution Agreement


cehbrecht:
@huard merged. Thanks :)

PR #383, opened 2018-08-17
fix update status doc (#377)

### LINKED ISSUE #332: status file does not update after an exception when running asynchronously
# Description

With the latest master, the xml status file does not update after an exception when running async. Here is an example process:

```
from pywps import Process
from pywps import LiteralOutput


class WPSraise(Process):
    def __init__(self):
        inputs = []

        outputs = [LiteralOutput('result',
                                 'Some result',
                                 data_type='float')]

        super(WPSraise, self).__init__(
            self._handler,
            identifier='raise',
            title='WPS raise',
            version='0.1',
            inputs=inputs,
            outputs=outputs,
            store_supported=True,
            status_supported=True)

    def _handler(self, request, response):
        raise RuntimeError("Some error")
        response.outputs['result'].data = 1.0
        return response
```

If I call (sync) using: http://localhost:8009/pywps?service=WPS&request=execute&version=1.0.0&identifier=raise&DataInputs=
everything is as before.
If I call (async) using: http://localhost:8009/pywps?service=WPS&request=execute&version=1.0.0&identifier=raise&storeExecuteResponse=true&status=true&DataInputs=
The xml file in statusLocation does not update, it is stuck on
```
<wps:ProcessAccepted>PyWPS Process raise accepted</wps:ProcessAccepted>
```
The exception still occurs in the apache error.log.

# Environment

- operating system: Ubuntu 16.04
- Python version: 2.7
- PyWPS version: https://github.com/geopython/pywps/archive/91da95ef846e4d9bce7cfd8016af467d2acfd1cb.zip
- source/distribution
 - [ X ] git clone
 - [ ] Debian
 - [ ] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [ X ] Apache/mod_wsgi
 - [ ] CGI
 - [ ] other (please specify):

### LINKED ISSUE #377: Progress of status document is not updated
# Description

When running a process in async mode the status document is not updated until the process has finished. At least the progress stays at 0 percent and there is no status message.

I can't say which PR has introduced this bug ... but my guess is #358. I need to investigate. 

# Steps to Reproduce

# Additional Information

cehbrecht:
# Overview

This PR fixes #377.

Changes:
* Fixed `update_status()` to use the current status instead of `WPS_STATUS.ACCEPTED`.
* Fixed execute response template: `status.message`.
* Reduced code duplication when generating json document for current status.

# Related Issue / Discussion

This PR probably also fixes #332.

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/response/execute.py
+++ b/pywps/response/execute.py
@@ -63,7 +63,7 @@ class ExecuteResponse(WPSResponse):
         """
         if status_percentage is None:
             status_percentage = self.status_percentage
-        self._update_status(WPS_STATUS.ACCEPTED, message, status_percentage)
+        self._update_status(self.status, message, status_percentage)
 
     def update_status_file(self, clean):
         # TODO: check if file/directory is still present, maybe deleted in mean time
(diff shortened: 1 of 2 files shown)


cehbrecht:
@eran-pinhas @jachym: I pull this PR in as well as a follow-up bug-fix to PR #373.

PR #385, opened 2018-08-20
set up requested mimetype of response

eran-pinhas:
set up the requested format as response.outputs[name].data_format

if the requested format is not in response.outputs[name].supported_formats - raise InvalidParameterValue error

# Contribution Agreement

- [ ] I'd like to contribute [feature X|bugfix Y|docs|something else] to PyWPS. I confirm that my 

### DIFF
--- a/pywps/app/Service.py
+++ b/pywps/app/Service.py
@@ -162,9 +162,6 @@ class Service(object):
                                     'MimeType ' + mimetype + ' not valid')
                             outpt.data_format = format[0]
 
-
-                    
-
         wps_response = process.execute(wps_request, uuid)
         return wps_response
 
--- a/pywps/app/Service.py
+++ b/pywps/app/Service.py
@@ -14,6 +14,7 @@ import pywps.configuration as config
 from pywps.exceptions import MissingParameterValue, NoApplicableCode, InvalidParameterValue, FileSizeExceeded, \
     StorageNotSupported, FileURLNotSupported
 from pywps.inout.inputs import ComplexInput, LiteralInput, BoundingBoxInput
+from pywps.inout.outputs import ComplexOutput
 from pywps.dblog import log_request, store_status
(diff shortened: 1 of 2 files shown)


eran-pinhas:
@cehbrecht - not an enhancement, it's bug (unless i got it all wrong). 

when the request specifies strictly what response mimetype it is expecting - the handler's response object is not being effected and therefore the handler is unable to answer with the right data

eran-pinhas:
btw, why does the coveralls check fails?

cehbrecht:
@eran-pinhas 

> btw, why does the coveralls check fails?

I suppose it needs to be configured to be less restrictive on the expected coverage. I would ignore it for now. 

@jachym Can you check the coverage config?

cehbrecht:
I have tested this PR with my Emu service using the following request:

http://localhost:5000/wps?version=1.0.0&service=WPS&request=Execute&identifier=wordcounter&DataInputs=text=@xlink:href=http://bird-house.github.io&ResponseDocument=output=@asReference=True@mimetype=application/json

When I choose an invalid like `mimetype=image/png` I get with PR an exception (as expected):
_MimeType image/png not valid_

It looks like the `mimetype` attribute is not case-insensitive ... probably another issue.

cehbrecht:
@eran-pinhas Thanks for fixing this. Maybe you could also extend the test-suite for this case? Otherwise I will pull it in.

cehbrecht:
> > It looks like the mimetype attribute is not case-insensitive ... probably another issue.

> I don't understand, what is the issue?

if you use `mimeType` instead of `mimetype` in the above request to the wordcounter process then the mimetype is not recognized. It is not related to this PR ... just noticed while testing.

eran-pinhas:
Done

eran-pinhas:
yes it is sound simpler, thank you

PR #386, opened 2018-08-21
Add traceback information to error message

huard:
# Overview

Simply adds traceback information to the error message to facilitate fixing bugs. Instead of a generic "No applicable error code, please check the error log", you get the full traceback. 

### DIFF
--- a/pywps/app/Service.py
+++ b/pywps/app/Service.py
@@ -414,7 +414,9 @@ class Service(object):
         except HTTPException as e:
             return NoApplicableCode(e.description, code=e.code)
         except Exception as e:
-            return NoApplicableCode("No applicable error code, please check error log.\n{0}".format(traceback.print_exception(*exc_info)), code=500)
+            msg = "No applicable error code, please check error log."
+            #logging.debug(msg + traceback.print_exception(*exc_info))
+            return NoApplicableCode(msg, code=500)
 
     @Request.application
     def __call__(self, http_request):
--- a/pywps/app/Service.py
+++ b/pywps/app/Service.py
@@ -21,6 +21,7 @@ from pywps.response.status import WPS_STATUS
 from collections import deque, OrderedDict
 import os
 import sys
+import traceback
 import uuid
 import copy
 import requests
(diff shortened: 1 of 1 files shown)


cehbrecht:
@huard I think returning just an error message without the traceback was changed recently ... not sure. Getting the traceback in development mode is quite useful. But probably in production we scare the user even more when we display a traceback. Could we make this optional? We might use the logging mode (DEBUG) for this ... or introduce a new `mode` parameter in the config.

Travis complains about pep8.

huard:
Maybe that's not the best fix, but my issue is that often pywps catches errors and returns an error message that doesn't provide information about where the bug is.  For example, if I put a 1/0 in outputs.py` ComplexOutput.json property, this is the error I get from the test suite:
```
======================================================================
FAIL: test_wfs (tests.test_ows.ExecuteTests)
----------------------------------------------------------------------
  File "/home/david/src/pywps/tests/test_ows.py", line 129, in test_wfs
    assert_response_success(resp)
  File "/home/david/src/pywps/pywps/tests.py", line 118, in assert_response_success
    assert resp.status_code == 200
AssertionError

```
To understand what's going on, I need to open a debug console, and print the resp.response to have a little bit more information, but it still lacks the actual traceback. 

I would argue that on the long run bugs will be easier to spot and fix with more informative and scary looking error messages. If someone sent you the error message above from a server that you don't have access to, there would be no way to guess at what the problem is. Is there a risk for tracebacks to leak sensitive information from the server ?

cehbrecht:
@huard did you revert the change? Well, I can't say if showing a traceback is a security issue. We could first make it work for us (easier debugging) and maybe add later an option to not expose this information if not wanted.

huard:
Yes, mistake on my end while fixing the codacy issue, but I'm not even sure my fix really works that well...

jachym:
Hi,

PyWPS should report only generic problem in the XML response, but full error message (full trace) should go to the log.

Yes, it would be security issue, having all the file names and line codes in the XML.

Issue #152, opened 2016-08-10
How to use OpenDAP with PyWPS?


cehbrecht:
i'm using PyWPS for processing of climate data. A major data format used here is NetCDF. In PyWPS one can use a complex type with mime-type `application/x-netcdf` to define input and output parameters for NetCDF files.

Another important way to access NetCDF files is with OpenDAP. A data service to access parts of the remote NetCDF file. The OGC WPS Best Practise paper recommends the mime-type `application/x-­ogc‐dods` for OpenDAP. If one would provide an OpenDAP url in a complex-type, PyWPS would try downloading it ... this is not the right behaviour, it should just pass it to the process. I could add a patch to fix this behaviour for OpenDAP mime-types.

A workaround would be to use a string parameter for the OpenDAP url. But then you loose the descriptive mime-type (which is used by generic wps clients). 

The Python NetCDF4 library can handle both NetCDF files and OpenDAP urls with the same initialisation code (`Dataset(file_or_url)`), so for a process it would not be necessary to have separate input parameters for OpenDAP and NetCDF.

Any recommendations on this topic?

Cheers,
Carsten

jachym:
Hi Carsten,

PyWPS is trying to handle the data inputs for you - and make sure, you are downloading right data, not too big, because you are operating with untrusted source. We try not to download data bigger than the configuration value (server -> maxsingleinputsize) and then validation can be applied (pywps.inout.formats.Format -> validate and mode attributes).

Having said that, I see your needs and it makes sense to me to address them. I can also see, having services around, like FME, it makes sense enable usage of such services and leave the input data handling and validating to them - to trusted 3rd party services.

Well, I think, we would need some complex approach to this topic, do you see any possible solution in 
https://github.com/geopython/pywps/blob/master/pywps/app/Service.py#L393 where the Reference data are downloaded?

cehbrecht:
@jachym 
i will have a look at it. I'm still on PyWPS 3.x ... so maybe i try it first there ... and hopefully switch to pywps 4.x soon.

cehbrecht:
We talked about this at the OGC coding-sprint in Bonn. I will make a small modification of the PyWPS 4.x code to make it possible to provide a netCDF file and a OpenDAP URL with the same `ComplexInput` parameter. netCDF and OpenDAP will have a different mimetype. Depending on the mimetype the PyWPS code can decide if it will download the file (netcdf) or just pass through the URL reference (OpenDAP) to the process.

[...]

PR #387, opened 2018-08-21
Implements explicit URL handling

huard:
# Overview

Adds a UrlHandler class. Previously, when users provided a URL to a ComplexInput, the file was automatically downloaded locally. With this PR, the link is instead stored in the url property and the file is only downloaded when the object's file, data or stream attributes are accessed. This allows processes using protocols such as opendap to use the url without downloading the file. 

Existing processes should not be affected because they would typically be accessing the file attribute in the _handler function. 

Note that the test coverage for ComplexOutput is somewhat thin. I suspect this PR could introduce bugs related to output properties. 

Tested with emu. 

# Related Issue / Discussion
Issue #352 

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -63,7 +63,7 @@ BoundingBoxData
 Request and response objects
 ----------------------------
 
-.. autodata:: pywps.response.status.STATUS
+.. autodata:: pywps.response.status.WPS_STATUS
     :annotation:
 
     Process status information
--- a/docs/migration.rst
+++ b/docs/migration.rst
@@ -54,7 +54,7 @@ you shall use first index of an input list::
 
 Inputs and outputs data manipulation
 ------------------------------------
-Btw, PyWPS Inputs do now have `file`, `data` and `stream` attributes. They are
+Btw, PyWPS Inputs do now have `file`, `data`, `url` and `stream` attributes. They are
 transparently converting one data-representation-type to another. You can read
 input data from file-like object using `stream` or get directly the data into
 variable with `input.data`. You can also save output data directly using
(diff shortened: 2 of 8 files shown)


cehbrecht:
I have double-checked this PR with [Emu](https://github.com/bird-house/emu) and it works. I have added a new process `ncmeta` to Emu which uses the `url` property introduced by this PR:
https://github.com/bird-house/emu/blob/61a151dc8ec7dfb4d71da548e1607ff6b4152a2d/emu/processes/wps_ncmeta.py#L54

cehbrecht:
@huard I have a comment to this PR and OpenDAP but I will add this to issue #152 ... for a next PR to support OpenDAP.

cehbrecht:
@huard Are you happy with the current state of this PR? I could pull it in so we can bring in the remaining steps with new PRs.

cehbrecht:
@huard travis is complaining about imports. You probably need to use Python 2/3 compatible imports:
https://github.com/geopython/pywps/blob/master/pywps/_compat.py

jachym:
Can we have the commits merged into one commit?

jachym:
I was running  integration tests at `pywps-demo` and it seems to be OK too. 

Do we need to update documentation too? @huard Just few notes to this topic?

jachym:
OK, once https://travis-ci.org/geopython/pywps/jobs/420109895 is fixed, we can merge

huard:
Done. Please review the changes to the docs. I've deleted a paragraph that I think was outdated (it said the file attribute returned a file object).

cehbrecht:
@jachym I pull it in. We can adapt things later if necessary.

cehbrecht:
@huard merged. Thanks :)

PR #394, opened 2018-08-28
NetCDF and OPeNDAP complex validators

huard:
# Overview

- Implement a complex validator for netCDF and OPeNDAP formats. 
- Adds a new DODS format for OPeNDAP mimetypes. 

# Related Issue / Discussion
#152 

# Additional Information
Entails a dependency on netCDF (optional). 

There is no change to the docs as this is not really discussed anywhere. 

# Contribution Agreement

### DIFF
--- a/pywps/app/Service.py
+++ b/pywps/app/Service.py
@@ -346,26 +346,6 @@ class Service(object):
         return self.call(http_request)
 
 
-def _openurl(inpt):
-    """use requests to open given href
-    """
-    data = None
-    href = inpt.get('href')
-
-    LOGGER.debug('Fetching URL %s', href)
-    if inpt.get('method') == 'POST':
-        if 'body' in inpt:
-            data = inpt.get('body')
-        elif 'bodyreference' in inpt:
-            data = requests.get(url=inpt.get('bodyreference')).text
-
-        reference_file = requests.post(url=href, data=data, stream=True)
-    else:
-        reference_file = requests.get(url=href, stream=True)
-
-    return reference_file
-
-
 def _build_input_file_name(href, workdir, extension=None):
     href = href or ''
     url_path = urlparse(href).path or ''
--- a/tests/test_execute.py
+++ b/tests/test_execute.py
@@ -9,7 +9,7 @@ import json
 import tempfile
 import os.path
 from pywps import Service, Process, LiteralOutput, LiteralInput,\
-    BoundingBoxOutput, BoundingBoxInput, Format, ComplexInput, ComplexOutput
+    BoundingBoxOutput, BoundingBoxInput, Format, ComplexInput, ComplexOutput, FORMATS
 from pywps.validator.base import emptyvalidator
 from pywps.validator.complexvalidator import validategml
 from pywps.validator.mode import MODE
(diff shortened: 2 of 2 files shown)


cehbrecht:
I'm using the validation mode with Emu. When I use an invalid input for OpenDAP (or NetCDF) the exception is not handled in the validator ... it should probably catch it and say the input is invalid, like here (opendap):

```
        try:
            from pywps.dependencies import netCDF4 as nc
            nc.Dataset(data_input.file)
            passed = True
        except ImportError:
            passed = False
        except Exception:   # or a more specific exception
            passed = False
```

cehbrecht:
I get an exception when I define the `default` attribute of a netCDF or OpenDAP `ComplexInput` with active validation (SIMPLE or STRICT):
`TypeError: expected str, bytes or os.PathLike object, not NoneType`

huard:
Are you using default_type=SOURCE_TYPE.URL ? 
Do you have an example in Emu I could run to reproduce the problem ?

huard:
The default handling question raises an issue I hadn't thought about. At the moment the default is set at the creation of ComplexInput, using the default source type, which is DATA. Whatever the default value, I can think of cases where the user request source type would be different from the default value. But changing the source type is not supported by IOHandler. 

I see two options, 1. add support for source type changes in IOHandler, 2. defer the setting of the default value until after the user request, when we know no values were provided and we need to actually set the object to its default value. I'd tend to go for the second option to avoid triggering actions (downloads, copies) for nothing.

cehbrecht:
From your explanation I also would prefer option 2. 

But changing the handling of the default value looks like quite some refactoring. Also we could then skip the `default_type` for `Inputs` ... would be an API change. Should this be a separate PR?

@jachym what is your opinion on this? The default will be set when a request comes in (if necessary) to make the `source_type` decision (File or URL) when we know the value.

cehbrecht:
@huard another thing I stumbled on. The setter for the default value does not know the URL type:
https://github.com/geopython/pywps/blob/1356504537711a1455a3b248de292000dd2d68e3/pywps/inout/basic.py#L193

When I add it then I run into more issues. The validator triggers a download to the `workdir` which does not exist ... we are at *definition* time, no process execution:
```
  File "/Users/pingu/Documents/GitHub/birdhouse/emu/emu/processes/wps_ncmeta.py", line 50, in __init__
    mode=MODE.STRICT),
  File "/Users/pingu/Documents/GitHub/birdhouse/pywps/pywps/inout/inputs.py", line 82, in __init__
    default=default, default_type=default_type)
  File "/Users/pingu/Documents/GitHub/birdhouse/pywps/pywps/inout/basic.py", line 851, in __init__
    self._set_default_value(default, default_type)
  File "/Users/pingu/Documents/GitHub/birdhouse/pywps/pywps/inout/basic.py", line 204, in _set_default_value
    self.url = value
  File "/Users/pingu/Documents/GitHub/birdhouse/pywps/pywps/inout/basic.py", line 261, in fset
    setattr(s, kls.prop, value)
  File "/Users/pingu/Documents/GitHub/birdhouse/pywps/pywps/inout/basic.py", line 416, in url
    self._check_valid()
  File "/Users/pingu/Documents/GitHub/birdhouse/pywps/pywps/inout/basic.py", line 157, in _check_valid
    _valid = validate(self, self.valid_mode)
  File "/Users/pingu/Documents/GitHub/birdhouse/pywps/pywps/validator/complexvalidator.py", line 243, in validatenetcdf
    name = data_input.file
  File "/Users/pingu/Documents/GitHub/birdhouse/pywps/pywps/inout/basic.py", line 424, in file
    self._file = self._build_file_name(href=self.url)
  File "/Users/pingu/Documents/GitHub/birdhouse/pywps/pywps/inout/basic.py", line 216, in _build_file_name
    input_file_name = os.path.join(self.workdir, file_name)
  File "/usr/local/anaconda3/envs/emu/lib/python3.6/posixpath.py", line 80, in join
    a = os.fspath(a)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

huard:
@cehbrecht Unless I'm wrong, it's a two line change to defer the evaluation of ComplexInput. I'll push the commit so you can take a look. I've also fixed the missing URL setting which I also ran into yesterday.

cehbrecht:
@huard It works :) If I understand it correctly you store the default in `_default` and in case of `ComplexInput` `_set_default_value` is only called when an execute request comes in.

Shall we set the `default_type` for `ComplexInput` to `SOURCE_TYPE.URL` by default? We mostly use URLs.

Still there are things to clean up and fix for defaults:
* default value is not shown in `DescribeProcess` for `ComplexInput` (template, json, ...).
* I doubt we need to explicitly set `default_type` ... it is either fixed (Literal, BBox) or we guess it from the value.
* Consistent handling of default for Literal, BBox and ComplexInput (default is set twice for Literal and BBox).

Shall we open another ticket for this?

huard:
Correct. 

I agree it would make sense to make changes to the default_type mechanisms, but I didn't want to break anything. 

I suppose it would be cleaner to make a separate PR with changes to default handling, but I can try to fix the describeprocess issue in this PR.

huard:
There is another issue with the json property: it accesses ComplexInput.file, which for a URL input would trigger a download. Suggestions ?

cehbrecht:
@huard I would not trigger `_set_default_value` in the json method. I have checked the WPS specs and it looks like the configured default value is not expected to be shown in the `DescribeProcess` response, just the formats (mimetypes):

http://www.opengeospatial.org/standards/wps (wps 1.0.0, table 21).

That was wishful thinking ;)

huard:
Ok, then I think we have a working version, with the caveat that a source_type clean-up remains to be done.

cehbrecht:
@jachym @huard I have opened a ticket (#395) for the issues with default values.

Issue #233, opened 2017-02-10
UTF-8 problems with output in Python 3


kvold:
# Description
If the output contains non-ascii characters Pywps (running in python3) is unable to encode them and throws the exception included below. 

Exception in the response:
<ows:Exception exceptionCode="NoApplicableCode" locater="None"><ows:ExceptionText>Process error: Process.py._run_process Line 227 'ascii' codec can't encode character '\xf8' in position 4600: ordinal not in range(128)

These are the lines (78, 79) in WPSResponse.py which causes the problem:
            with open(self.process.status_location, 'w') as f:
                f.write(etree.tostring(doc, pretty_print=True, encoding='utf-8').decode('utf-8'))

The problem can be fixed by encoding the bytes explicitly, though this might cause problems for python2 (I haven't checked):
            with open(self.process.status_location, 'wb') as f:
                f.write(etree.tostring(doc, pretty_print=True, encoding='utf-8'))

# Environment
- operating system: Centos 7
- Python version: 3.4.5
- PyWPS version: 4.0.0
- source/distribution
 - [X] git clone
 - [ ] Debian
 - [ ] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [X] Apache/mod_wsgi
 - [ ] CGI
 - [ ] other (please specify): 

# Steps to Reproduce
Running Pywps in Python 3 and write non-ascii characters to output.

# Additional Information
My terminal uses utf-8 as its character encoding.

From the pywps-log:
[ERROR] file=/usr/local/wps/src/pywps/pywps/app/Process.py line=250 module=Process function=_run_process Process error: Process.py._run_process Line 227 'ascii' codec can't encode character '\xf8' in position 4600: ordinal not in range(128)

From the Apache-log:
[Fri Feb 10 10:06:24.643446 2017] [wsgi:error] [pid 11339] Traceback (most recent call last):
[Fri Feb 10 10:06:24.643793 2017] [wsgi:error] [pid 11339]   File "/usr/local/wps/src/pywps/pywps/app/Process.py", line 227, in _run_process
[Fri Feb 10 10:06:24.643804 2017] [wsgi:error] [pid 11339]     100, STATUS.DONE_STATUS, clean=self.async)
[Fri Feb 10 10:06:24.644008 2017] [wsgi:error] [pid 11339]   File "/usr/local/wps/src/pywps/pywps/app/WPSResponse.py", line 70, in update_status
[...]

PR #400, opened 2018-09-07
Added unicode output in tests

huard:
# Overview

Just added unicode strings in ComplexOutputTest to verify it is handled properly. 

# Related Issue / Discussion
#233 

### DIFF
--- a/tests/test_inout.py
+++ b/tests/test_inout.py
@@ -1,3 +1,5 @@
+# -*- coding: utf-8 -*-
+
 """Unit tests for IOs
 """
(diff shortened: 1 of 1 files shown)


cehbrecht:
@huard 
The test looks simple.

Don't know if issue #233 is really solved ... it mentions a WPS exception. At the moment there is only a standard exception returned ("... check the logs.") ... maybe that is why we don't get this error :)

cehbrecht:
@huard pulled in. Thanks :)

Issue #410, opened 2018-09-12
Format information is missing from describe process when output is asReference


huard:
# Description
Mime types, encoding and schema information is missing for outputs marked asReference. Is this the desired behavior? Also, the template main.xml lacks the asReference information. 

# Environment

- PyWPS version: master
- source/distribution
 - [x] git clone

cehbrecht:
@huard I think that is fixed with your PR #411.

PR #411, opened 2018-09-12
Include format information for "referenced" ComplexData

huard:
# Overview
Include more information about outputs in describeprocess. 

# Related Issue / Discussion
#410 


cehbrecht:
@huard can you point me in the wps 1.0.0 spec where this is described?
http://www.opengeospatial.org/standards/wps

The spec is no fun to read :)

I have found `asReference` only in table 51 which belongs to the `Execute` response document.

The description of complex output for `DescribeProcess` response is in table 21.

huard:
You're right, my mistake.

jachym:
@huard I think, asReference does not belong to describeprocess - only execute response

huard:
Right, the PR title was misleading. Fixed it.

PR #419, opened 2018-11-07
Fixes #418

### LINKED ISSUE #418: LiteralInput min_occurs, max_occurs attributes always set to 1
# Description
There is a bug in basic.py LiteralInput class, when it initialises its parent BasicIO, it sets the min_occurs, max_occurs to 1 instead of using the passed in parameter.  The same goes for the metadata parameter. 
This results in all process LiteralInputs being mandatory and only accepting 1 value when more than 1 are passed into the process.  So for example, a process defined with an input parameter with max_occurs=9 which should allow for 9 possible values will only make use of 1. That is because in service.py (create_literal_inputs), line 216 a deque is created using the LiteralInput's max_occurs parameter to limit the size for that parameter.  This also means that the describeProcess  output for any process is incorrect, max_occurs and min_occurs are always set to 1.
# Environment

- operating system: Ubuntu 18.04.1 LTS
- Python version: 3.6
- PyWPS version: 4
- source/distribution
 - [x] git clone
 - [ ] Debian
 - [x ] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [ x] Apache/mod_wsgi
 - [ ] CGI
 - [ ] other (please specify): 

# Steps to Reproduce
Create a process, set min_occurs or max occurs > 1
# Additional Information

sbar13:
# Overview

This PR fixes issue #418.

# Related Issue / Discussion

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/inout/basic.py
+++ b/pywps/inout/basic.py
@@ -677,7 +677,7 @@ class LiteralInput(BasicIO, BasicLiteral, SimpleHandler):
                  min_occurs=1, max_occurs=1, metadata=[],
                  default=None, default_type=SOURCE_TYPE.DATA):
         BasicIO.__init__(self, identifier, title, abstract, keywords,
-                         min_occurs=1, max_occurs=1, metadata=[])
+                         min_occurs, max_occurs, metadata)
         BasicLiteral.__init__(self, data_type, uoms)
         SimpleHandler.__init__(self, workdir, data_type, mode=mode)
 


cehbrecht:
Tests fail due to pep8 checks ... but not caused by this PR.

cehbrecht:
@sbar13 merged. Thanks :)

Issue #425, opened 2018-12-11
Adding "debug" mode to pywps ... and at least keep the workdirs


cehbrecht:
# Description

In the current pywps (master) the workdir is removed at the end of a process, no matter the status of the process (failed, success). Check the `clean` function:
https://github.com/geopython/pywps/blob/fd45020a491764228f503b13829b60ebc1e1544b/pywps/app/Process.py#L267

Is is good for production but during development it can be very helpful to keep the workdir to have a chance to see what has gone wrong.

We should introduce a *debug* or *testing* mode in pywps to keep the workdir. Probably in all cases (process failure and success).

# Environment

- operating system:
- Python version:
- PyWPS version: pre 4.2.0
- source/distribution
 - [ ] git clone
 - [ ] Debian
 - [ ] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [ ] Apache/mod_wsgi
 - [ ] CGI
 - [ ] other (please specify): 

# Steps to Reproduce

# Additional Information

cehbrecht:
I think the easiest fix is just to add a config option for *debug*/*testing* mode and check in the `clean` function if it is set.

cehbrecht:
Do we want to have this option in the server section?
https://github.com/geopython/pywps/blob/fd45020a491764228f503b13829b60ebc1e1544b/pywps/configuration.py#L76

```
[server]
testmode=true
```

cehbrecht:
fixed by PR #426.

PR #426, opened 2018-12-11
added testing mode (#425)

cehbrecht:
# Overview

This PR adds a testing mode to PyWPS.

* added `testmode` configuration parameter as boolean flag.
* updated docs for `testmode`.
* disabled removal of process `workdir` when `testmode=true`. 

# Related Issue / Discussion

#425.

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/app/Process.py
+++ b/pywps/app/Process.py
@@ -267,15 +267,20 @@ class Process(object):
     def clean(self):
         """Clean the process working dir and other temporary files
         """
-        LOGGER.info("Removing temporary working directory: %s" % self.workdir)
-        try:
-            if os.path.isdir(self.workdir):
-                shutil.rmtree(self.workdir)
-            if self._grass_mapset and os.path.isdir(self._grass_mapset):
-                LOGGER.info("Removing temporary GRASS GIS mapset: %s" % self._grass_mapset)
-                shutil.rmtree(self._grass_mapset)
-        except Exception as err:
-            LOGGER.error('Unable to remove directory: %s', err)
+        testmode = config.get_config_value('server', 'testmode')
+        if testmode is True:
+            LOGGER.warning('PyWPS is running in testing mode. Workdir is not removed: {}'.format(
+                self.workdir))
+        else:
+            LOGGER.info("Removing temporary working directory: %s" % self.workdir)
+            try:
+                if os.path.isdir(self.workdir):
+                    shutil.rmtree(self.workdir)
+                if self._grass_mapset and os.path.isdir(self._grass_mapset):
+                    LOGGER.info("Removing temporary GRASS GIS mapset: %s" % self._grass_mapset)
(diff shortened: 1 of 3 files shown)


cehbrecht:
@huard Please review. Add changes to the PR branch if you like.

pywps/app/Process.py
```diff
@@ -267,15 +267,20 @@ def launch_next_process(self):
...
-        try:
-            if os.path.isdir(self.workdir):
-                shutil.rmtree(self.workdir)
-            if self._grass_mapset and os.path.isdir(self._grass_mapset):
-                LOGGER.info("Removing temporary GRASS GIS mapset: %s" % self._grass_mapset)
-                shutil.rmtree(self._grass_mapset)
-        except Exception as err:
-            LOGGER.error('Unable to remove directory: %s', err)
+        testmode = config.get_config_value('server', 'testmode')
+        if testmode is True:
```
tomkralidis: Better written as `if testmode:`

pywps/app/Process.py
```diff
@@ -267,15 +267,20 @@ def launch_next_process(self):
...
+        else:
+            LOGGER.info("Removing temporary working directory: %s" % self.workdir)
+            try:
+                if os.path.isdir(self.workdir):
+                    shutil.rmtree(self.workdir)
+                if self._grass_mapset and os.path.isdir(self._grass_mapset):
+                    LOGGER.info("Removing temporary GRASS GIS mapset: %s" % self._grass_mapset)
+                    shutil.rmtree(self._grass_mapset)
+            except Exception as err:
+                LOGGER.error('Unable to remove directory: %s', err)
```
tomkralidis: What's the canonical way of formatting strings in `logging` statements espeically in a Python 3 context?  Is `%s` ok?  Would be valuable to know for other projects as well.

pywps/app/Process.py
cehbrecht: done.

pywps/app/Process.py
cehbrecht: Can't say what the official recommendation is. Change to traditional style and updated log message.

tomkralidis: If the `{}` / `.format()` approach works, can we go with that to be more future proof?

cehbrecht: I think this is a topic for a next PR. But I haven't found an official statement concerning the logging. But I like the `format()` approach ;)

cehbrecht:
@huard ... I would pick `noclean` then.

pywps/app/Process.py
cehbrecht: I consider this solved.

pywps/app/Process.py
cehbrecht: I have opened an issue for this question #427.

cehbrecht:
using `cleantempdir` option now.

Issue #412, opened 2018-09-19
ComplexOutput as data does not always work


huard:
# Description

When as_reference is False in a ComplexOutput, but the content is provided with the file attribute, no data is included in the response, only the reference. 

When the content is provided by the data attribute, then the data is sent, but includes HTML escape characters and CDATA statements. Is this expected? 

# Environment

- operating system: ubuntu
- Python version: 3.6
- PyWPS version: master
- source/distribution
 - [x] git clone
 - [ ] Debian
 - [ ] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [ ] Apache/mod_wsgi
 - [ ] CGI
 - [ ] other (please specify): 

# Steps to Reproduce
Tweak emu/emu/processes/wps_multiple_outputs.py to set as_reference to False for both ComplexOutputs. Then run emu/tests/test_wps_multiple_outputs.py and inspect the output. 

# Additional Information
It seems these cases are not exercised by the test suite.

cehbrecht:
fixed by PR #431.

PR #431, opened 2018-12-13
Bug fix 412

huard:
# Overview

Fixes issue with ComplexOutput and as_reference. 

When ComplexOutput was set by a file, it forced as_reference to True. 
Binary data was output as string instead of base64. 
Also, the raw CDATA content had extra white space. 

# Related Issue / Discussion
https://github.com/geopython/pywps/issues/412


jachym:
thanks @cehbrecht @lazaa32 and @huard

PR #432, opened 2018-12-14
Bug fix 412

lazaa32:
# Overview
Fixes issue with ComplexOutput and as_reference.

When ComplexOutput was set by a file, it forced as_reference to True.
Binary data was output as string instead of base64.
Also, the raw CDATA content had extra white space.
CDATA tag was used even for XML-based output.

# Related Issue / Discussion
#412 

# Contribution Agreement

### DIFF
--- a/pywps/inout/basic.py
+++ b/pywps/inout/basic.py
@@ -416,7 +416,6 @@ class UrlHandler(FileHandler):
         self._reset_cache()
         self._url = value
         self._check_valid()
-        self.as_reference = True
 
     @property
     def file(self):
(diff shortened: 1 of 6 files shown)


cehbrecht:
@lazaa32 thanks for this PR :) 

It has a python 3.x issue ... see travis tests.

cehbrecht:
... well, even with `six` it needs to be handled different, failed now on 2.7
```
  File "tests/test_inout.py", line 338, in test_base64
    self.complex_out.data = six.binary_type(self.data, 'utf-8')
TypeError: str() takes at most 1 argument (2 given)
```

huard:
Fixed that on https://github.com/huard/pywps/tree/bug_fix_412

@lazaa32 I removed the spacing in the main.xml template because when converting back the data, these spaces end up being part of the response. It makes the xml ugly, but I'm not aware of a workaround.

cehbrecht:
closing in favor of PR #431.

PR #440, opened 2018-12-19
Do not call Session.close_all

elemoine:
# Overview

Do not call Session.close_all as this will close all the Sessions that exist in memory, including the sessions that are specific to the application and unrelated to PyPWS.

And here the closing of sessions is already handled by _LAST_SESSION.close(), so we can just remove the call to _SESSION_MAKER.close_all.

# Related Issue / Discussion

None

# Additional Information

None. The overview covers it all.

# Contribution Agreement

### DIFF
--- a/pywps/dblog.py
+++ b/pywps/dblog.py
@@ -155,7 +155,6 @@ def get_session():
         _LAST_SESSION.close()
 
     if _SESSION_MAKER:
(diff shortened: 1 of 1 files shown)


cehbrecht:
@elemoine done. Thanks :)

elemoine:
Thanks to you!

PR #443, opened 2018-12-20
Docker extension

lazaa32:
# Overview
Added new ``Container.py`` class for Docker extension.
The extension works so far only for pywps-flask demo server. See [pywps-flask PR](https://github.com/geopython/pywps-flask/pull/46).

# Related Issue / Discussion

# Additional Information

# Contribution Agreement


requirements.txt
```diff
@@ -8,3 +8,4 @@ python-dateutil
 requests
 SQLAlchemy
 werkzeug
+docker
```
cehbrecht: can we move `docker` to the optional dependencies?

pywps/app/Process.py
```diff
@@ -172,20 +173,32 @@ def _execute_process(self, async, wps_request, wps_response):
             if running >= maxparallel and maxparallel != -1:
                 raise ServerBusy('Maximum number of parallel running processes reached. Please try later.')
             wps_response._update_status(WPS_STATUS.ACCEPTED, u"PyWPS Request accepted", 0)
-            wps_response = self._run_process(wps_request, wps_response)
+            wps_response = self._run_sync(wps_request, wps_response)
```
cehbrecht: why is it called `run_sync`? docker will only be run in async mode.

cehbrecht (changes requested):
Just left some comments. There also pep8 issues.

pywps/app/Process.py
lazaa32: We discussed that with @jachym and we concluded that the docker should be run even in sync mode once the ``mode`` parameter is set to ``docker`` in config file.

fmigneault:
Closing since very outdated.

PR #444, opened 2018-12-20
Io tests

huard:
# Overview

Added tests for the raw data round trip. 

I've managed to get what I think is acceptable behavior for file formats that have example files in the test suite. 

Now if I try to send a txt file including reserved XML characters, it breaks and I haven't been able to fix that. My guess is that it requires some fiddling on the OWSLib side to add a CDATA tag then remove it in pywps. 

# Related Issue / Discussion

# Additional Information

# Contribution Agreement


huard:
I just realized that the `safe` filter in the templates means that the content is marked as safe, thus won't be escaped. Is that what is wanted here, or we'd rather use the `escape` filter to make sure that HTML characters don't mess with the XML ? That requires the client to unescape the output.

cehbrecht:
I copy+paste a few things found in the WPS 1.0.0 spec.

Table 47: ComplexData input as embedded data
> The complex value is embedded here as part of the ComplexData element, in the mimeType, encoding, and schema indicated by the first three parameters if they exist, or by the relevant defaults.

Examples:
* http://schemas.opengis.net/wps/1.0.0/examples/51_wpsExecute_request-extended_ResponseDocument.xml
* http://schemas.opengis.net/wps/1.0.0/examples/53_wpsExecute_request_ComplexValue.xml

So ... it is not encoded ... or the encoding is described (utf-8).

cehbrecht:
For outputs I have chapter 10.3.1 (Execute response parameters):
> If storage of a ComplexData output has not been requested via the ―asReference‖ attribute (see Table 51), the ExecuteResponse document will contain that output. For ComplexData such as images included in the ExecuteResponse document will be encoded. 

They have not given an example but I suppose it should be treated like the inputs.

Example with ref:
http://schemas.opengis.net/wps/1.0.0/examples/62_wpsExecute_response.xml

cehbrecht:
In case of binary data (images, pdf, netcdf, ....) we need to encode with base64 and set the attribute:
```
encoding="base64"
```

In all other cases (xml, json, text, ...) we don't need to encode and may set the attribute:
```
encoding="UTF-8"
```

The client needs to check the `encoding` attribute and handle it appropriately.

When will it go wrong?

huard:
Would that mean setting the encoding to utf-8 for all non-binary pywps.FORMATS ?
I'm not sure that solves the problem of non-xml safe characters, but I'm always confused with encoding.

cehbrecht:
> I just realized that the safe filter in the templates means that the content is marked as safe, [..]

The jinja docs says we only use the `safe` filter when the templates are encoded automatically. We can also encode manually:
http://jinja.pocoo.org/docs/2.10/templates/

huard:
I think we'll need to bring back the CDATA tag for non-xml, non-base64 content.

cehbrecht:
> ... but I'm always confused with encoding.

Well, me too :) Unfortunately the docs and examples are a bit thin. Can we get the simplest case working for us that seems to be inline with the spec? This PR already looks like it ...

cehbrecht:
Just a link for cdata:
https://www.tutorialspoint.com/xml/xml_cdata_sections.htm

So, for non-base64 and non-xml we would use cdata (text, json, ....)? The encoding is still then "UTF-8". Does the client side (owslib, xml parser) handle this?

huard:
The PR chokes on non-xml safe characters. So that needs to be solved.

huard:
In owslib.wps https://github.com/geopython/OWSLib/blob/dcbc8db9a8eafa715a0469317bf23ea07de17326/owslib/wps.py#L1331
there is a strip method applied on the ComplexData content. Is that necessary ? I think it would be best not to temper with the content.

pywps/inout/inputs.py
```diff
@@ -134,22 +134,26 @@ def _json_data(self, data):
...
+                # Otherwise we assume all other formats are unsafe and need to be enclosed in a CDATA tag.
                 elif isinstance(self.data, bytes):
-                    data["data"] = self.data.encode('utf8')
+                    out = self.data.encode(self.data_format.encoding or 'utf-8')
 
                 else:
-                    data["data"] = self.data
+                    out = self.data
+
+                data["data"] = u'<![CDATA[{}]]>'.format(out)
```
cehbrecht: This puts `cdata` tag around base64 encoded data as well? I think we don't need this for base64 ... but probably it is harmless.

cehbrecht:
> In owslib.wps https://github.com/geopython/OWSLib/blob/dcbc8db9a8eafa715a0469317bf23ea07de17326/owslib/wps.py#L1331
there is a strip method applied on the ComplexData content. Is that necessary ? I think it would be best not to temper with the content.

Looking at "blame" in `owslib.wps` this is old code. `strip` removes whitespace at the beginning and end for the string. I don't think it will do any harm to us.

cehbrecht:
> In owslib.wps https://github.com/geopython/OWSLib/blob/dcbc8db9a8eafa715a0469317bf23ea07de17326/owslib/wps.py#L1331
there is a strip method applied on the ComplexData content. Is that necessary ? I think it would be best not to temper with the content.

Looking at "blame" in `owslib.wps` this is old code. `strip` removes whitespace at the beginning and end for the string. I don't think it will do any harm to us.

pywps/inout/inputs.py
huard: Correct, mistake on my end.

huard:
I still think it would be desirable to preserve the enclosed content with as little interference as possible, so that sending a file and retrieving it back results in no modification to the file.

pywps/inout/inputs.py
cehbrecht: done.

cehbrecht:
@huard Let me know when ready. I can merge then.

huard:
strip: I've built a workaround in the test suite, so not a priority. Let's leave it like that until a real problem comes up. 
Please merge.

cehbrecht:
@huard done :)

Issue #442, opened 2018-12-19
Status code 200 when error in execution handler


eran-pinhas:
# Description

Hey, 

I believe there's an issue with the status code. when the handler itself raises an error the Process.py catches it (lines below) sets the wps_request as failed (attached below) but then continues like no error occurred. As result the status code of the HTTP request is 200 and should be 500

https://github.com/geopython/pywps/blob/ec1d90f7424598a662f68c28e4de4452e2bdeaf3/pywps/app/Process.py#L232

I can think of 2 solutions:
* re-raise an NoApplicableCode at [Process.py#L232](https://github.com/geopython/pywps/blob/master/pywps/app/Process.py#L232)
* before the `return response` [(Service.py#L326)](https://github.com/geopython/pywps/blob/master/pywps/app/Service.py#326) check whether the response is failed one and if so raise an NoApplicableCode that will be catched at [Service.py#L335](https://github.com/geopython/pywps/blob/master/pywps/app/Service.py#L335)

What do you guys think? @cehbrecht @jachym 

# Environment

- operating system: ubuntu 18
- Python version: 3.6
- PyWPS version: latest from github
- source/distribution: git clone
- web server: Flask

eran-pinhas:
@cehbrecht what do you think? I would be happy to help fixing the issue. just need your opinion on how to do it.

cehbrecht:
We should use the Exceptions (HTTP status codes) according to the WPS protocol:
http://www.opengeospatial.org/standards/wps

For WPS 1.0.0 see Table 62 in the spec.

Here is a list of the WPS 2.0.0 protocol:
http://docs.opengeospatial.org/is/14-065/14-065.html

I'm not sure if we need to raise an exception when the process has failed. It will be reported with an status document ... and using status code=200. 

The exceptions available look like to report that there is something wrong with my WPS service. But to report process failures you just use a status document.

@jachym You are more familiar with the WPS protocol.

cehbrecht:
Execute Exceptions in WPS 2.0.0:
http://docs.opengeospatial.org/is/14-065/14-065.html#61

cehbrecht:
WPS 1.0.0:
> [..] If a process fails for some reason, the implementation raises an error and places it in the exception report included in this ProcessFailed structure.

It sounds confusing to me ... but the `ProcessFailed` structure is part of the status report ... not part of a WPS exception.

jachym:
Hi,

[...]

PR #446, opened 2019-01-01
Reraised NoApplicableCode with status 500 at handler catch

eran-pinhas:
As discussed at #442 

- I have already previously agreed to the PyWPS Contributions and Licensing Guidelines

### DIFF
--- a/pywps/app/Process.py
+++ b/pywps/app/Process.py
@@ -231,6 +231,8 @@ class Process(object):
                 msg = 'Process failed, please check server error log'
             wps_response._update_status(WPS_STATUS.FAILED, msg, 100)
 
+            raise NoApplicableCode('Process failed, please check server error log', code=500)
+
         finally:
             # The run of the next pending request if finished here, weather or not it successfull
             self.launch_next_process()


cehbrecht:
@eran-pinhas @jachym technically this works. But I can't decide how the WPS protocol wants to have it. See my comments in #442.

cehbrecht:
@eran-pinhas @jachym @davidcaron I would like to solve this PR. I still don't know how to decide but from tests with [Emu (dry-run process)](https://emu.readthedocs.io/en/latest/processes.html#simpledryrun) I'm in favor of not merging.

Currently when I run a wrong wps request (missing parameter, etc) I get the following exception type:
```
<ows:ExceptionReport xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd" version="1.0.0">
  <ows:Exception exceptionCode="NoApplicableCode" locator="">
    <ows:ExceptionText>Process failed, please check server error log</ows:ExceptionText>
  </ows:Exception>
</ows:ExceptionReport>
```

When something goes wrong within my process (for example using the dry-run exception in emu example), I get an exception in a status document:
```
<wps:ExecuteResponse xmlns:wps="http://www.opengis.net/wps/1.0.0" xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/wps/1.0.0 ../wpsExecute_response.xsd"
  service="WPS" version="1.0.0" xml:lang="en-US" serviceInstance="http://localhost:5000/wps?request=GetCapabilities&amp;amp;service=WPS" statusLocation="">
  <wps:Process wps:processVersion="1.5">
    <ows:Identifier>hello</ows:Identifier>
    <ows:Title>Say Hello</ows:Title>
    <ows:Abstract>Just says a friendly Hello. Returns a literal string output with Hello plus the inputed name.</ows:Abstract>
  </wps:Process>
  <wps:Status creationTime="2019-01-08T18:23:27Z">
    <wps:ProcessFailed>
      <wps:ExceptionReport>
        <ows:Exception exceptionCode="NoApplicableCode" locator="None">
          <ows:ExceptionText>Process failed, please check server error log</ows:ExceptionText>
        </ows:Exception>
      </wps:ExceptionReport>
    </wps:ProcessFailed>
  </wps:Status>
</wps:ExecuteResponse>

davidcaron:
@cehbrecht I think I agree with you, and the specification is a bit confusing.

Section 10.3.3 Execute exceptions: 
> When a WPS server encounters an error while performing an Execute operation, it shall
return an exception report message as specified in Subclause 8 of [OGC 06-121r3].

The response format of this section corresponds to what we would get if this PR gets merged. It seems to me that these exceptions are all raised before starting the process execution. (missing or invalid parameters, storage and queue problems, etc.) These exceptions seem to be more about "we couldn't start processing for some reason".

Section 10.3.1 Execute response parameters (Note d in Table 55)
> If a process fails for some reason, the implementation raises an error and places it in the exception report included in this ProcessFailed structure.

The responses in table 55 are about the processing status, and a failed status is still a valid status.

So I think that the specs say that this PR should not be merged... but at the same time, I agree that a 200 status code containing an error message is not ideal.

eran-pinhas:
Hey guys, thank for your time trying to understand the WPS spec with me. We actually used your way of interpretation and didn't use the status code. And because I'm the only one cared for it I think we can close it and there will be no hard feelings 😁

jachym:
thano you, @eran-pinhas for your contribution anyway

PR #455, opened 2019-03-05
Improve queue race conditions

### LINKED ISSUE #450: Response document of a stored job is not updated
# Description

I make a asynchronous process that works perfectly when the job is alone (parallelprocesses=1). So, to test the maxprocesses=10, i send two requests to my server. The second job is stored and a document has built. But, when the first job is finished and the second is released, the document response of second wasn't updated, but the database sqlite and log file are updated and files are produced in work directory. The response document remains stuck with <wps:ProcessAccepted percentCompleted="0">PyWPS Process waterFlowSimulation accepted</wps:ProcessAccepted>.

I don't understand the problem because the response document of the first job is updated <wps:ProcessSucceeded>PyWPS Process Estimation  waterFlowSimulation finished</wps:ProcessSucceeded>.

 Do you have any idea where the problem may come from ?

# Environment

- operating system : Debian GNU/Linux 9.5
- Python version: 2.7.15rc1
- PyWPS version: 4.2.1
- source/distribution
 - [ ] git clone
 - [ ] Debian
 - [ ] PyPI
 - [x] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [x] Apache/mod_wsgi
 - [ ] CGI
 - [ ] other (please specify): 

# Steps to Reproduce
1/ Set parallelprocesses=1 and maxprocesses=10
2/ Execute 2 process at same time with storeExecuteResponse=true (async) and get there statusLocation
3/ The response document of the process stored hasn't updated, but the process is done (if files are created, you can find them on the server). The response document of the first process is updated.

# Additional Information
I tried to change the database sqlite with postgres, but the problem is the same. The process has updated inside pywps requests table, but not inside the response document.

davidcaron:
# Overview

When I tested sending lots of small requests to pywps:

- Multiple threads couldn't write to the database at the same time (a global session was shared between threads)
- Workers would take the same request twice from the stored_requests table
- Unfinished requests would pile up and being considered as running, even though an error happened

I found that making Pywps completely threadsafe and multi-process safe is challenging. When I started using locks, the forked Pywps workers would copy the opened lock to the forked process, leading to a deadlock.

Another solution would be to [spawn](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods) processes instead of forking them. But this is Python 3.4+ only.

So I fixed a few things so that race conditions are less likely to happen.

# Related Issue / Discussion

Also, I think 24f48dd fixes #450

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/app/Process.py
+++ b/pywps/app/Process.py
@@ -270,6 +270,7 @@ class Process(object):
             process._set_uuid(uuid)
             process.async = True
             new_wps_response = ExecuteResponse(new_wps_request, process=process, uuid=uuid)
+            new_wps_response.store_status_file = True
             process._run_async(new_wps_request, new_wps_response)
         except Exception as e:
             LOGGER.exception("Could not run stored process. {}".format(e))
--- a/pywps/app/Process.py
+++ b/pywps/app/Process.py
@@ -147,8 +147,8 @@ class Process(object):
         """
 
         maxparallel = int(config.get_config_value('server', 'parallelprocesses'))
-        running = dblog.get_running().count()
-        stored = dblog.get_stored().count()
+
+        running, stored = dblog.get_process_counts()
 
         # async
         if async:
--- a/pywps/app/Process.py
+++ b/pywps/app/Process.py
@@ -154,15 +154,19 @@ class Process(object):
         if async:
 
             # run immedietly
+            LOGGER.debug("Running processes: {} of {} allowed parallelprocesses".format(running, maxparallel))
+            LOGGER.debug("Stored processes: {}".format(stored))
+
             if running < maxparallel or maxparallel == -1:
                 wps_response._update_status(WPS_STATUS.ACCEPTED, u"PyWPS Request accepted", 0)
+                LOGGER.debug("Accepted request {}".format(self.uuid))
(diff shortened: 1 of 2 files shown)


cehbrecht:
@davidcaron is there a difference if we use sqlite or postgres? Would that influence the current PR?

huard:
I can't review this PR but these issues have caused us some trouble lately. We'd be really glad if this could be fixed.

davidcaron:
@cehbrecht I tested both postgres and sqlite, and I found that postgres handles concurrent connections better than sqlite... which isn't surprising, but sqlite should at least do the job.

This PR improves connections using both sqlite and postgres, but in my tests, I found that sqlite had more problems than postgres, and I couldn't pinpoint what caused sqlite to crash. I got file corruption error messages from sqlite.

Keep in mind that I was sending short bursts of 4-5 requests at the same time, and I never got any issue when the request has time to start processing before receiving another one.

cehbrecht:
@davidcaron @jachym I don't have much insight in the database handling but since it improves the situation I will merge it.

cehbrecht:
@davidcaron done. Thanks :)

PR #456, opened 2019-03-05
Fix deserialization

davidcaron:
# Overview

I had trouble with the serialization-deserialization of a ComplexInput being a url link to a file. 

One of the issue is that the `file` property should be saved for a ComplexInput that has a `prop` attribute of `url`. 

I'm not 100% sure that it should be fixed this way, but it fixes the case of the `url` type and I don't think that other types are impacted. 

Also, when deserializing the WPSRequest, I believe the inputs should be from `pywps.inout.inputs` instead of `pywps.inout.basic`. The docstring mentions that the latter are abstract types.

# Related Issue / Discussion

# Additional Information

# Contribution Agreement


cehbrecht:
@davidcaron Could you add a test to show the "serialization-deserialization" of a ComplexInput?

pywps/inout/inputs.py
```diff
@@ -99,7 +98,8 @@ def json(self):
             'data_format': self.data_format.json,
             'asreference': self.as_reference,
             'supported_formats': [frmt.json for frmt in self.supported_formats],
-            'file': self.file if self.prop == 'file' else None,
+            'file': self.file,
```
huard: The issue with this is that it's going to force a download of the URL locally. Look at the URLHandler `file` property getter in `inout/basic.py`. 
The url is already stored under the `href` key.

davidcaron: would this work then?
` self.file if self.prop in ['file', 'url'] else None`

For the 'url' case (`UrlHandler`), the file is already downloaded at this point, and when the request is deserialized, no other download takes place... so we need the reference to the file property.

davidcaron: The file property of `UrlHandler` starts like this:

```python

    @property
    def file(self):
        if self._file is not None:
            return self._file

        self._file = self._build_file_name(href=self.url)
        .....
```

huard: If the file is downloaded by URLHandler, then there's a bug.

In theory, the file is downloaded when the `file` property's getter is called. For the URLHandler, this should not happen unless explicitly requested by the Process handler. The point of the URLHandler is precisely to avoid downloading the file locally and let the process handler deal with data access. Our use case for this was netCDF files accessed through the opendap protocol. Users can send an opendap link to a process and have the process extract and process a few grid cells without having to download the entire file to the PyWPS server. 

Your fix would still force a download, which we want to avoid if `prop == 'url'`.

davidcaron:
@cehbrecht sure, I will add tests when I get some time.

davidcaron:
I moved json serialization methods for the inputs from WPSRequest.json to inout.inputs so that they are close to each other. I put them in inout.inputs instead of inout.basic because ComplexInput.json was already there and json methods for outputs are in inout.outputs, not in inout.basic.

I had to fix a circular dependency between inout.formats and validator.complexvalidator. In all cases where Format is instantiated, the `validate` property is set. So this shouldn't break existing code. I prefer this solution to having a delayed import...

I fixed other things I found along the way, making this PR larger that it was originally.

@huard , please check https://github.com/geopython/pywps/pull/456/commits/ced88fbcb4bd082de9c8aa91f2158e5a67e91013 I think this would cover all cases of ComplexInput.

davidcaron:
Codacy is giving me "Dangerous default value [] as argument" but this happens a lot elsewhere. I agree this should be changed, but maybe not in this PR...

huard:
Does the Emu test suite also pass with those changes ? I've catched some issues in the past that we're missed by the pywps test suite.

cehbrecht:
I wasn't really aware of this issue ... thought it would be just coding style. We should open an issue and fix it in a new PR, maybe like this:
```
def __init__(self, some_list=None):
    self.some_list = some_list or []
```

> Codacy is giving me "Dangerous default value [] as argument" but this happens a lot elsewhere. I agree this should be changed, but maybe not in this PR...
> 
> I think this should really be changed, because this can lead to very hard to find bugs:
> 
> ```python
> class Test:
>     def __init__(self, some_list=[]):
>         self.some_list = some_list
> 
> 
> t1 = Test()
> t2 = Test()
> 
> t2.some_list.append(1)
> print(t1.some_list)  # prints [1]
> ```

cehbrecht:
@davidcaron @huard I have checked this PR branch with Emu and there are issues at least with the bbox:
```
'BoundingBoxOutput' object has no attribute 'json'
```

Just run the `getcapabilities` request.

> Does the Emu test suite also pass with those changes ? I've catched some issues in the past that we're missed by the pywps test suite.

davidcaron:
@cehbrecht I think this should fix it. I had not touched the code for the outputs, but `inputs.BoundingBoxOutput` was inheriting from `basic.BBoxInput`.

cehbrecht:
@huard @davidcaron If I heare a "go" by you and no complains by others I can merge :)

davidcaron:
I'm done, feel free to @ me if you notice any related bug after the merge.

davidcaron:
Sorry, I just realized the Metadata class needed serialization methods also

cehbrecht:
@davidcaron done. Thanks :)

huard:
@davidcaron I believe this PR introduced a bug with the describeprocess response. Below is an excerpt of a LiteralData element. Note that the value is a string representation of a property, instead of the property value. 
 
```
<LiteralData xmlns:wps="http://www.opengis.net/wps/1.0.0" xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
                <ows:DataType ows:reference="http://www.w3.org/TR/xmlschema-2/#float">float</ows:DataType>
                <ows:AllowedValues>
                    <ows:Value>&lt;property object at 0x7febfb256ea8&gt;</ows:Value>
                </ows:AllowedValues>
                <DefaultValue>2.0</DefaultValue>
                </LiteralData>
```

PR #457, opened 2019-03-08
remove a sleep timer that is only useful for async requests

davidcaron:
But annoying for sync requests and tests. The purpose of this timer is to
cover a case when a request finishes more or less at the same
time another is stored. The stored request is not started until another
request ends.

Introduced in #455 that was recently merged.


cehbrecht:
@davidcaron Haven't noticed this issue. I will merge.

Issue #397, opened 2018-08-31
make pywps python 3.7 compatible


cehbrecht:
# Description

Python 3.7 is available and it comes with some changes. For example:

> W606  'async' and 'await' are reserved keywords starting with Python 3.7 

```
INFO:conda_build.build:Packaging pywps-4.1.14-py37_0
compiling .pyc files...
  File "lib/python3.7/site-packages/pywps/app/Process.py", line 99
    self.async = False
             ^
SyntaxError: invalid syntax
```

# Environment

# Steps to Reproduce

# Additional Information

See also ticket in OWSLib: https://github.com/geopython/OWSLib/issues/455

Zeitsperre:
Ran into this problem today. Glad that this is already being looked into. :)

cehbrecht:
we probably need to fix owslib first:
https://github.com/geopython/OWSLib/issues/515

czhang11:
Same issue. Python 3.7.

cehbrecht:
we might want to change the naming to mode to preserve consistency with owslib,

PR #462, opened 2019-03-30
Rename async to async_ for Python 3.7 compatibility.

sebastic:
`async` became a reserved keyword in Python 3.7:
> Backwards incompatible syntax changes:
>
> * [async](https://docs.python.org/3/reference/compound_stmts.html#async) and [await](https://docs.python.org/3/reference/expressions.html#await) are now reserved keywords.

https://docs.python.org/3/whatsnew/3.7.html

Fixes: #397

### DIFF
--- a/pywps/app/Process.py
+++ b/pywps/app/Process.py
@@ -97,7 +97,7 @@ class Process(object):
 
     def execute(self, wps_request, uuid):
         self._set_uuid(uuid)
-        self.async = False
+        self.async_ = False
         response_cls = get_response("execute")
         wps_response = response_cls(wps_request, process=self, uuid=self.uuid)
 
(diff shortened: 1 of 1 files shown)


cehbrecht:
I'm not sure if we want parameter names ending with `_`. In OWSLib we changed it to use `mode` instead:
https://github.com/geopython/OWSLib/pull/464

@jachym also suggested to use a longer name `asynchron`.

@tomkralidis @huard Do you have an opinion on this?

tomkralidis:
Perhaps `mode` to be consistent with other efforts in OWSLib?

sebastic:
Appending an underscore is standard to resolve conflicts with reserved keywords. If you want to entirely rename the variable that fine too, and you can close this issue.

huard:
I agree we might want to change the naming to `mode` to preserve consistency with owslib, but I also think it's useful to have an immediate fix for this and create another issue to switch the name.

PR #464, opened 2019-04-02
fix for `allowed_values == AnyValue` in templates

davidcaron:
# Overview

The `any_value` parameter was left out of the `LiteralInput` serialization, but the xml template response depends on it.

# Related Issue / Discussion

#456 

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/inout/literaltypes.py
+++ b/pywps/inout/literaltypes.py
@@ -83,8 +83,6 @@ class AllowedValue(object):
                  minval=None, maxval=None, spacing=None,
                  range_closure=RANGECLOSURETYPE.CLOSED):
 
-        AnyValue.__init__(self)
-
         self.allowed_type = allowed_type
         self.value = value
         self.minval = minval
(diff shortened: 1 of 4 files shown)


pywps/inout/literaltypes.py
```diff
@@ -341,7 +335,9 @@ def make_allowedvalues(allowed_values):
 
     for value in allowed_values:
 
-        if isinstance(value, (AllowedValue, AnyValue, NoValue, ValuesReference)):
+        if value in ALLOWED_VALUES_TYPES:
+            new_allowedvalues.append(value())
+        elif isinstance(value, ALLOWED_VALUES_TYPES):
```
cehbrecht: should this be `isinstance(value, ALLOWED_VALUE_TYPE)`?

davidcaron: Do you mean changing the name `ALLOWED_VALUES_TYPES` to `ALLOWED_VALUE_TYPE`?

davidcaron: If not, the 2 checks are necessary, (`value in ALLOWED_VALUES_TYPES` and `isinstance(value, ALLOWED_VALUES_TYPES)`) because the defaut value for `LiteralInput.allowed_values` is `AnyValue` which is the class directly, instead of an instance.

cehbrecht: but `ALLOWED_VALUES_TYPES` it defined like this:
```
ALLOWED_VALUES_TYPES = (AllowedValue, AnyValue, NoValue, ValuesReference)
```

while there is also something like this:
```
from pywps.validator.allowed_value import ALLOWEDVALUETYPE
```

davidcaron: Ah, I see... maybe I could rename ALLOWED_VALUES_TYPES to ALLOWED_VALUES_CLASSES? or maybe reference the classes by their specific names directly instead of creating this tuple...

It doesn't serve the same purpose, ALLOWEDVALUETYPE is just a container for the constants 'value' and 'range'.

cehbrecht: ... I didn't know that  `isinstance` can also check against a tuple of classes. So, you can keep it like this ... maybe just a comment to avoid the confusion.

cehbrecht:
@davidcaron merged. Thanks :)

PR #465, opened 2019-04-05
fix sphinx build

cehbrecht:
# Overview

Sphinx build was failing due to this errors:
* `AttributeError` exception having no `message` on Python 3.6.
* `AnyValue` missing `value`. 

Note, `AnyValue` is not really implemented yet.

# Related Issue / Discussion

# Additional Information

# Contribution Agreement


cehbrecht:
@huard Please, check if this fix is ok.

PR #466, opened 2019-04-05
Add metalink support to facilitate outputs with multiple files

### LINKED ISSUE #298: New Feature: return multiple output files via a reference document (json)
# Description

In our projects we have quite often the case where a process produces multiple output files ... not knowing the number in advance and it could be many (10, 100). The WPS protocol allows us only to define an output parameter which returns a single output file reference. Unlike for the WPS input parameter one can not  define a multiplicity (min_occurs, max_occurs).

Our current workaround is not so nice ... we tar/zip the output files and return a reference to the archive.

A nicer approach could be as follows:

* output files are written to a folder in the storage (file storage currently).
* output folder name could be a uuid.
* a json document is generated with URL references to all files in the output storage.
* the json document could include metadata, which is usually part of an output parameter (title, abstract, mime-type).
* this json document is returned via a "normal" WPS output parameter (reference document, "virtual link").
* to make the special use of this json reference document recognizable by WPS clients, one could either have a special mime-type (are these kinds of documents already used somewhere?) or ... one could use `ows:Metadata` to indicate the semantic of the output parameter.

What do you think? 

I would like to collect some feedback first ... the implementation could be made by us.

# Additional Information

COWS WPS has implemented something similar ... it returns an XML document with references including additional metadata.

An example (under development) for a json reference document:
https://github.com/Ouranosinc/flyingpigeon/blob/spatial-op-wfs-opendap/flyingpigeon/handler_common.py#L114

Another example, which uses pywps code ... but it is not working that way ... but you may get the idea:
https://github.com/bird-house/emu/blob/master/emu/processes/wps_multiple_outputs.py

huard:
# Overview
Fix #298 

Changes : 
- Classes to facilitate the creation of metalink files: 
  - MetaFile class
  - MetaLink and MetaLink4 classes (v3.0, v4.0)
- METALINK and META4 mime type formats
- xml validator
- schemas for strict xml validation 
- documentation section 
- basic tests

This PR was also tested with the Emu test server and the Birdy test client. 

# Contribution Agreement


cehbrecht:
@tomkralidis @jachym do you have any objections to pull this in? I can do a squash merge. We have already tested it. See the updated docs in this PR (`docs/processes.rst`) for motivation. See also:
https://www.wikiwand.com/en/Metalink

cehbrecht:
@huard done :)

PR #468, opened 2019-04-10
Fix 406 user exceptions

### LINKED ISSUE #406: User defined exceptions?
# Description

I'm coming back to the discussion in PR #386. Currently PyWPS shows only a common WPS exception when an error has occurred ("_Please check the logs ..._"), introduced in PR #357.

A WPS exception is a way to communicate to the user ... we like to give some hints when something goes wrong. Though we don't want (or should not) show the raw traceback which is probably not understandable for the end-user and a potential security risk.

Could we introduce a new "User Exception"? 

In the process implementation we could through this user exception intentionally do give the user some hints of what might have gone wrong ("_No data found for given selection parameter ..._"). We could add restrictions to the user exception type, like only 300 chars, no suspicious characters (no tracebacks), ... In all other cases the common exception will be thrown ("_Check the logs_").

@jachym @huard 
What do you think?

# Environment

- operating system:
- Python version:
- PyWPS version:
- source/distribution
 - [ ] git clone
 - [ ] Debian
 - [ ] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [ ] Apache/mod_wsgi
 - [ ] CGI
 - [ ] other (please specify): 

# Steps to Reproduce

# Additional Information

See discussion in PRs #386, #357.

Also we need to check the Python encodings: #233, #400.

cehbrecht:
# Overview

This PR introduces new exceptions which can be raised from the process code. 

Unlike to common exceptions, a message send using the new `ProcessError` exception will be returned to the user (wps client) via an OGC exception. The message text is validated before sending (not too long, no suspicious chars). If the message is not valid then a default message will be used.

This `ProcessError` exception is intentionally used to provide some "meaningful" information to the user of what went wrong. Otherwise we always get the message "sorry, something went wrong, please check the logs", which is of not much help to end-users. 

This PR fixes #406.

# Related Issue / Discussion

#406.

# Additional Information

See an example in Emu:
https://github.com/bird-house/emu/blob/758563fdee2f468c0e94392192475e3d3ca5037e/emu/processes/wps_error.py#L38

# Contribution Agreement

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -10,6 +10,10 @@ Process
 
 .. autoclass:: Process
 
+Exceptions you can raise in the process implementation to show a user-friendly error message.
+
+.. autoclass:: pywps.app.exceptions.ProcessError
+
 Inputs and outputs
 ==================
 
--- a/docs/process.rst
+++ b/docs/process.rst
@@ -388,6 +388,42 @@ Example process
 .. literalinclude:: metalinkprocess.py
    :language: python
 
+Process Exceptions
+==================
+
+Any uncatched exception in the process execution will be handled by PyWPS and reported
+to the WPS client using an `ows:Exception`. PyWPS will only log the traceback and report
+a common error message like:
+
+    *Process failed, please check server error log.*
+
+This sparse error message is used to avoid security issues by providing internal
+service information in an uncontrolled way.
+
+But in some cases you want to provide a user-friendly error message to give the user a hint of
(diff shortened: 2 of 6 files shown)


cehbrecht:
@huard Please have a look at it again.

I need to add some documentation to this PR. But first, I would like to know if you accept the implementation.

cehbrecht:
updated the docs and added an example.

cehbrecht:
@tomkralidis merged. Thanks for review :)

Issue #405, opened 2018-09-10
BoundingBox implementation is broken


cehbrecht:
# Description

I tried to use the `bbox` and I run into the following exception:

```
2018-09-10 14:11:06,149] [INFO] line=306 module=Service Request: execute
2018-09-10 14:11:06,160] [ERROR] line=45 module=exceptions Exception: code: 500, locator: No applicable error code, please check error log., description: 
Traceback (most recent call last):
  File "/usr/local/anaconda3/envs/emu/lib/python3.6/site-packages/pywps/app/Service.py", line 332, in call
    raise e
  File "/usr/local/anaconda3/envs/emu/lib/python3.6/site-packages/pywps/app/Service.py", line 325, in call
    request_uuid
  File "/usr/local/anaconda3/envs/emu/lib/python3.6/site-packages/pywps/app/Service.py", line 81, in execute
    return self._parse_and_execute(process, wps_request, uuid)
  File "/usr/local/anaconda3/envs/emu/lib/python3.6/site-packages/pywps/app/Service.py", line 130, in _parse_and_execute
    inpt, request_inputs)
  File "/usr/local/anaconda3/envs/emu/lib/python3.6/site-packages/pywps/app/Service.py", line 273, in create_bbox_inputs
    newinpt.data = [datainput.minx, datainput.miny,
AttributeError: 'dict' object has no attribute 'minx'

```
# Environment

- operating system:
- Python version:
- PyWPS version: pre 4.2 on master
- source/distribution
 - [ ] git clone
 - [ ] Debian
 - [ ] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [ ] Apache/mod_wsgi
 - [ ] CGI
 - [ ] other (please specify): 

# Steps to Reproduce

I'm using the bbox process in [Emu](https://github.com/bird-house/emu/blob/master/emu/processes/wps_bbox.py).

See bbox test:
https://github.com/bird-house/emu/blob/master/tests/test_wps_bbox.py

# Additional Information

cehbrecht:
Fixed by PR #551

PR #472, opened 2019-05-03
Fixed a bug in the execute xml jinja template

tommygod3:
# Overview
Small issue that was causing the rendering of a result XML to crash when using a bbox input. 

# Contribution Agreement

- [X] I'd like to contribute this bugfix to PyWPS. I confirm that my contributions to PyWPS will be compatible with the PyWPS license guidelines at the time of contribution.


cehbrecht:
This is partial fix for #405. It occurred when using `lineage=true` and the inputs were rendered in a execute response document.

cehbrecht:
@tommygod3 thanks :)

PR #474, opened 2019-05-10
fix for a LiteralInput with a default value of 0

davidcaron:
# Overview

When a LiteralInput is configured to have a default value of 0, or anything that is not None but 'falsy', The default is absent in the DescribeProcess template.

# Related Issue / Discussion

https://github.com/bird-house/birdy/issues/120

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/templates/1.0.0/describe/literal.xml
+++ b/pywps/templates/1.0.0/describe/literal.xml
@@ -32,6 +32,6 @@
                     {% endfor %}
                 </ows:AllowedValues>
                 {% endif %}
-                {% if put.data %}
+                {% if put.data is not none %}
                 <DefaultValue>{{ put.data }}</DefaultValue>
                 {% endif %}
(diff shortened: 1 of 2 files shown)


pywps/templates/1.0.0/describe/literal.xml
```diff
@@ -32,6 +32,6 @@
                     {% endfor %}
                 </ows:AllowedValues>
                 {% endif %}
-                {% if put.data %}
+                {% if put.data is not none %}
```
cehbrecht: shouldn't this be `is not None`?

davidcaron: Yes, that's what I thought first also. Jinja uses test functions, not straight python evaluation. See: https://github.com/pallets/jinja/issues/520

cehbrecht: ok.

cehbrecht:
@davidcaron merged. Thanks :)

PR #475, opened 2019-08-06
Fix - set store file encoding to utf8

eran-pinhas:
# Contribution Agreement

### DIFF
--- a/pywps/inout/basic.py
+++ b/pywps/inout/basic.py
@@ -297,7 +297,7 @@ class FileHandler(IOHandler):
     def data(self):
         """Read file and return content."""
         if self._data is None:
-            with open(self.file, mode=self._openmode()) as fh:
+            with open(self.file, mode=self._openmode(), encoding='utf8') as fh:
                 self._data = fh.read()
         return self._data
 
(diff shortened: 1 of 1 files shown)


pywps/inout/basic.py
```diff
@@ -297,7 +297,9 @@ def file(self, value):
     def data(self):
         """Read file and return content."""
         if self._data is None:
-            with open(self.file, mode=self._openmode()) as fh:
+            openmode = self._openmode()
+            kwargs = {} if 'b' in openmode else {'encoding': 'utf8'}
```
cehbrecht: `encoding` is only available in Python 3.x? We still have Python 2.7 compatibility. Using `io.open` might work for 2.7 and 3.x.

pywps/inout/basic.py
```diff
@@ -5,6 +5,7 @@
 
 from pywps._compat import text_type, StringIO
 import os
+from io import open
```
cehbrecht: maybe use it in the form:
```
import io
io.open()
```

and add a comment like "PY2 compatibility". When we skip PY2 we can simplify it again.

cehbrecht: see here:
https://stackoverflow.com/questions/40809901/python-2-7-io-open-use-from-io-import-open-or-io-open#40809945

cehbrecht: ... I will add this in a follow-up PR.

cehbrecht:
@eran-pinhas merged. Thanks :) I will change the `io.open` import in another PR.

PR #483, opened 2019-08-19
using io.open

cehbrecht:
# Overview

This PR is updating PR #475 and uses `io.open`.

# Related Issue / Discussion

#475

# Additional Information

https://stackoverflow.com/questions/40809901/python-2-7-io-open-use-from-io-import-open-or-io-open#40809945

# Contribution Agreement

### DIFF
--- a/pywps/inout/basic.py
+++ b/pywps/inout/basic.py
@@ -5,7 +5,7 @@
 
 from pywps._compat import text_type, StringIO
 import os
-from io import open
+import io  # TODO: using io.open for PY2 compatibility
 import shutil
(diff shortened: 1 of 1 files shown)


cehbrecht:
Looks like this approach leads into more PY2 compatibility troubles:
```
File "pywps/inout/basic.py", line 387, in file

    fh.write(self.data)

TypeError: write() argument 1 must be unicode, not str
```

cehbrecht:
skipping this PR ... it becomes obsolete when using Python 3 only.

Issue #477, opened 2019-08-07
Shall we skip Python 2.7 support?


cehbrecht:
# Description

Supporting only Python 3.x makes the code easier to maintain. In `OWSLib` we have already done this step.

What is your opinion?

ldesousa:
I agree. Python 2.7 is now being dropped from most Linux distributions, pretty soon no one will use it any more.

Let us keep the discussion open for now and have a PSC vote later in the month. Hopefully there will be quorum during FOSS4G.

jachym:
Motion: get rid of Python 2 support for PyWPS 4.3 (future 4.4) branch

- [ ] @cehbrecht 
- [ ] @jachym
- [ ] @ldesousa 
- [ ] @tomkralidis 
- [ ] @jonas-eberle 
- [ ] @jorgejesus

jachym:
done, thank you all

PR #485, opened 2019-08-20
added from_json to ValuesReference

cehbrecht:
# Overview

This PR adds `ValuesReferences.from_json` which is need by the scheduler for the json dump/load. 
It updates the test cases for processing and fixes the `Literalnput.json` for data types like `time`.

# Related Issue / Discussion

#482

# Additional Information

With this PR the scheduler extension works for the Emu `inout` process:

# Contribution Agreement

### DIFF
--- a/pywps/inout/inputs.py
+++ b/pywps/inout/inputs.py
@@ -290,14 +290,14 @@ class LiteralInput(basic.LiteralInput):
             'workdir': self.workdir,
             'allowed_values': [value.json for value in self.allowed_values],
             'any_value': self.any_value,
-            'values_reference': self.values_reference,
             'mode': self.valid_mode,
             'min_occurs': self.min_occurs,
             'max_occurs': self.max_occurs,
-
             # other values not set in the constructor
-            'data': self.data,
+            'data': str(self.data),
         }
+        if self.values_reference:
+            data['values_reference'] = self.values_reference.json
         if self.uoms:
             data["uoms"] = [uom.json for uom in self.uoms]
         if self.uom:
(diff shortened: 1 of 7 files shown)


cehbrecht:
@davidcaron would you like to have a look at the code changes? I take care of the scheduler extension ... it is not so easy to set-up the test environment.

cehbrecht:
Ok, still some issues with `LiteralInput.json`:
```
File "/home/travis/build/geopython/pywps/tests/test_inout.py", line 637, in test_json_out_time

    self.assertEqual(out['data'], datetime.time(12, 30, 0), 'time set')

AssertionError: '12:30:00' != datetime.time(12, 30) : time set
```

pywps/inout/inputs.py
```diff
@@ -313,7 +313,10 @@ def from_json(cls, json_input):
             elif allowed_value['type'] == 'novalue':
                 allowed_values.append(NoValue())
             elif allowed_value['type'] == 'valuesreference':
-                allowed_values.append(ValuesReference())
+                allowed_values.append(ValuesReference(
```
davidcaron: I guess you could do : `allowed_values.append(ValuesReference.from_json(allowed_value))`

davidcaron:
As far as I know, the serialization-deserialization of `values_reference` should work like you did it.

Now, like you pointed out, I'm not sure how to guess the format of `str(data)` when deserializing...

pywps/inout/inputs.py
cehbrecht: done. Also for `AllowedValues`.

cehbrecht:
> Now, like you pointed out, I'm not sure how to guess the format of `str(data)` when deserializing...

I stumbled upon it. So, there are already `convert` methods to go from `str to object`. Just by luck the opposite (from object to str) worked for `datetime` with a `str(data)`. It might not be enough ... but it looks like I need to update the failing tests. The json should not contain a Python object.

Thoughts?

cehbrecht:
I have updated the `ConverterTests` with checks like this:
```
self.assertEqual(convert_time(str(datetime.time(12, 0, 0))),
                         datetime.time(12, 0, 0))
```

cehbrecht:
@davidcaron I would like to fix the tests. Are you ok with this?

davidcaron:
Ok, I just found what I was looking for, I didn't see it before. When deserializing, the type is taken from `data_type` and passed to the `convert` function when the `data` property is set.

So as far as I know this should work as expected.

cehbrecht:
@davidcaron I have fixed the tests. I was lazy and skipped the Python 2.7 failures (unicode issues). We plan to skip Python 2.7 (#477) ... so maybe not worth to fix it.

Are you ok with this PR?

cehbrecht:
@davidcaron merged. Thanks for review :)

Issue #448, opened 2019-01-17
Clean up after process crash


huard:
# Description

The database entries (pywps_stored_requests, pywp_requests) for requests are not decremented when processes crash. After some time the server reaches the maximum number of parallel processes allowed and rejects all new requests, even though no process is active. 

Is there a way to catch the crashes and remove those processes from the count to avoid this issue ?

omerlevko:
hi,
@cehbrecht  - do you have any idea about how to solve this? i will implement it

cehbrecht:
@huard Probably this will be solved by the watchdog daemon in PR @497.

PR #497, opened 2019-10-10
Daemon

jachym:
This PR introduces new way of async process execution. Async processes are stored to database.

User also needs to start `pywps_daemon.py` script, which will be running as background process and check the database for incoming requests and execute them (using the `processing.Job` mechanism)

Besides that, some smaller changes are introduced:

* Removing tests of Python 2
* Removing async execution from Process class
* Adding more pythonish class attribute setters for uuid
* Adding Json representation of process to stored_processes
* Adding dbglog tests

# Contribution Agreement

### DIFF
--- a/bin/pywps_daemon.py
+++ b/bin/pywps_daemon.py
@@ -1,0 +1,144 @@
+#!/usr/bin/env python3
+
+from daemon import DaemonContext
+import sys
+import time
+import logging
+
+from pywps import dblog
+from pywps.app.WPSRequest import WPSRequest
+from pywps.app.Service import Service
+import pywps.processing
+import pywps.configuration as config
+import json
+import lockfile
+import os
+import signal
+
+# Configure logging
+LOGGER = logging.getLogger("PYWPS")
+
+
+def watchdog():
+
+    global LOGGER
+    config.load_configuration()
+
+    LOGGER.setLevel(getattr(logging, config.get_config_value('logging', 'level')))
+
+    max_time = int(config.get_config_value('daemon', 'pause'))
+    maxparallel = int(config.get_config_value('server', 'parallelprocesses'))
+
+    while True:
+        # Logging errors and exceptions
+        try:
+            running, stored = dblog.get_process_counts()
+            LOGGER.info('PyWPS daemon: {} running processes {} stored requests'.format(
+                running, stored))
+
+            while (running < maxparallel or maxparallel == -1) and stored > 0:
+                launch_process()
+                running, stored = dblog.get_process_counts()
+
+        except Exception as e:
+            LOGGER.exception("PyWPS daemon failed: {}".format(str(e)))
+
+        # The daemon will repeat your tasks according to this variable
+        # it's in second so 60 is 1 minute, 3600 is 1 hour, etc.
+        time.sleep(max_time)
+
+
+def launch_process():
+    """Look at the queue of async process, if the queue is not empty launch
+    the next pending request.
+    """
+    try:
+        from pywps.processing.job import Job, JobLauncher
+
+        LOGGER.debug("Checking for stored requests")
+
+        stored_request = dblog.pop_first_stored()
+        if not stored_request:
+            LOGGER.debug("No stored request found, sleeping")
+            return
+
+        value = {
+            'process': json.loads(stored_request.process.decode("utf-8")),
+            'wps_request': json.loads(stored_request.request.decode("utf-8"))
+        }
+        job = Job.from_json(value)
+
+        processing_process = pywps.processing.Process(
+            process=job.process,
+            wps_request=job.wps_request,
+            wps_response=job.wps_response)
+        processing_process.start()
+
+    except Exception as e:
+        LOGGER.exception("Could not run stored process. {}".format(e))
+        raise e
+
+
+def start(pidfile):
+
+    with DaemonContext(pidfile=pidfile) as context:
+        context.signal_map = {
+            # signal.SIGTERM: program_cleanup,
+            signal.SIGHUP: 'terminate',
+            # signal.SIGUSR1: reload_program_config,
+        }
+
+        watchdog()
+
+
+if __name__ == "__main__":
+
+    config.load_configuration()
+    pidfile = config.get_config_value("daemon", "pidfile")
+    lockfile = lockfile.FileLock(pidfile)
+    print("PIDFile: ", pidfile)
+    print("LOCKFile: ", lockfile)
+
+    if len(sys.argv) == 2:
+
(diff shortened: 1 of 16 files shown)


jachym:
Hm .. probably shell script will be needed for better implementation of the `start` `stop` and `restart` methods :-|

cehbrecht:
Running a test with Emu WPS. sync requests work as expected. Running an async execute request I get the following error:
```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) table pywps_stored_requests has no column named process
[SQL: INSERT INTO pywps_stored_requests (uuid, request, process) VALUES (?, ?, ?)]
```

We are missing an upgrade mechanism for the database?

In [twitcher](https://twitcher.readthedocs.io/en/latest/installation.html#initialize-database) we are using [alembic](https://pypi.org/project/alembic/).

cehbrecht:
@jachym @huard @davidcaron 
In birdhouse deployment we currently use Werkzeug to start the WPS service for development:
https://github.com/bird-house/cookiecutter-birdhouse/blob/7a1259731d1d532b719bcc3fcbe036578fc95fa7/%7B%7Bcookiecutter.project_repo_name%7D%7D/%7B%7Bcookiecutter.project_slug%7D%7D/cli.py#L134

In production mode we use supervisor deployed by Ansible:
https://github.com/bird-house/ansible-wps-playbook/blob/master/roles/pywps/templates/supervisor.conf.j2

When we merge this PR our installations will break! We need to update our service start scripts.

If we accept this PR it needs to go into a 4.4.0 release (together with dropping Python 2.x).

cehbrecht:
@jachym 
This PR with separation of the watchdog code to poll jobs from the database makes the code look cleaner. It might also make the service less error-prone.

Could the watchdog also be a cronjob? So we would just need to add a cronjob (configuration task) instead of running an additional service?

We probably need to fix the database upgrade.

bin/pywps_daemon.py
```diff
@@ -0,0 +1,144 @@
+#!/usr/bin/env python3
+
+from daemon import DaemonContext
```
cehbrecht: requirement for [python-daemon](https://pypi.org/project/python-daemon/) is missing.

huard:
Question: Is this PR intended to fix #495 ? #448 ?

pywps/dblog.py
```diff
@@ -170,13 +173,16 @@ def get_session():
                 engine = sqlalchemy.create_engine(database,
                                                   echo=echo,
                                                   connect_args={'check_same_thread': False},
-                                                  poolclass=NullPool)
```
davidcaron: Do you really need a `QueuePool` here? I believe you definitely need to keep `NullPool`, because the multiprocessing forks, and can lead to open/broken connections. I believe there will basically be no performance impact for this.

davidcaron:
> 
> 
> @jachym @huard @davidcaron
> In birdhouse deployment we currently use Werkzeug to start the WPS service for development:
> https://github.com/bird-house/cookiecutter-birdhouse/blob/7a1259731d1d532b719bcc3fcbe036578fc95fa7/%7B%7Bcookiecutter.project_repo_name%7D%7D/%7B%7Bcookiecutter.project_slug%7D%7D/cli.py#L134
> 
> In production mode we use supervisor deployed by Ansible:
> https://github.com/bird-house/ansible-wps-playbook/blob/master/roles/pywps/templates/supervisor.conf.j2
> 
> When we merge this PR our installations will break! We need to update our service start scripts.

Thank you for the head up.

My 2 cents: 

I think having a separate process to launch jobs is a good idea and will make the queue more resilient. It's closer to something like a celery task queue.

We deploy in containers, so the daemon will be in its own container, and a separate entrypoint command for the daemon is much better than a cron job in this case (1 process = 1 container).

pywps/dblog.py
davidcaron: See: https://github.com/geopython/pywps/commit/203ae6dc83f46d783852a3b72b693a5d2301f453

cehbrecht:
Tested with several emu sleep process. It works when I start the watchdog ... it fills the queue and runs until all jobs are finished:
`pywps_daemon.py watchdog`

... but the daemon mode ends immediately:
`pywps_daemon.py start`

cehbrecht:
@jachym @davidcaron 
> We deploy in containers, so the daemon will be in its own container, and a separate entrypoint command for the daemon is much better than a cron job in this case (1 process = 1 container).

Could we provide both, cron-job and daemon service? So we can support several deployment scenarios. I think I would feel safer with a cron-job in the ansible/supervisor scenario. For the docker/container and also for the Werkzeug/demo mode we might want a daemon service.

davidcaron:
> Could we provide both, cron-job and daemon service? So we can support several deployment scenarios. I think I would feel safer with a cron-job in the ansible/supervisor scenario. For the docker/container and also for the Werkzeug/demo mode we might want a daemon service.

Sure! I didn't mean one or the other. I agree the cron-job would be better with ansible outside of containters.

pywps/app/Process.py
```diff
@@ -7,7 +7,6 @@
 import os
```
jorgejesus: Note, this import
https://github.com/geopython/pywps/blob/bde355145462d0767f359e4a8a6303d0606e618d/pywps/app/Process.py#L23

`from pywps._compat import PY2` this doesnt  seem to be necessary any more

pywps/dblog.py
jorgejesus: @davidcaron  `NullPool` will have a penalties performance if you use PostGRESQL, here should be `NullPool` to avoid thread problems

cehbrecht:
@jachym @davidcaron @huard I want to work on this "watchdog" PR the next days to fix the job queue handling. I will open a new PR for this ... it will include the code of this one. I want to move code to a new `watchdog` module and move the command-line interface of our birdhouse template to pywps:
https://github.com/bird-house/cookiecutter-birdhouse/issues/52

pywps/dblog.py
jachym: I believe, this is just for the sqlite, as @jorgejesus is writing, we are constantly hitting threads problem with sqlite (any best-practice solution would be great)

bin/pywps_daemon.py
jachym: done

pywps/app/Process.py
jachym: done

jachym:
@cehbrecht shall we close this PR and continue in #505  (?)

cehbrecht:
> @cehbrecht shall we close this PR and continue in #505 (?)

Yes ... I'm closing this PR. Maybe we should have worked on dev branch.

PR #499, opened 2019-10-24
Check data is defined in literal template

### LINKED ISSUE #498: Literal template return DefaultValue when data is not defined
# Description

The describeProcess response includes an empty DefaultValue field, which OWSLib 0.18 chokes on. 

# Environment

- operating system: ubuntu
- Python version: 3.6
- PyWPS version: 4.2.2

huard:
# Overview
I think this started causing issues with owslib 0.18. 

# Related Issue / Discussion
Fixes #498 

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/templates/1.0.0/describe/literal.xml
+++ b/pywps/templates/1.0.0/describe/literal.xml
@@ -32,6 +32,6 @@
                     {% endfor %}
                 </ows:AllowedValues>
                 {% endif %}
-                {% if put.data is not none %}
+                {% if put.data is defined and put.data is not none %}
                 <DefaultValue>{{ put.data }}</DefaultValue>
                 {% endif %}


cehbrecht:
@huard merged. Thanks.

tlvu:
Could a conda and pypi release be made to make this fix available? Thanks.
________________________________

cehbrecht:
@tlvu does it help to use the master branch in your deployment?
https://github.com/bird-house/emu/blob/a307fdf5998623c2b9bf5f5e08fe908c4b4bc03a/environment.yml#L18

We don't have so many fixes since the last release ... 

@jachym @tomkralidis 
Would you mind to do a 4.2.3 release or should we wait?

jachym:
If @tlvu is ok with usage of master branch, I would wait till there are more PRs in the branch.

But if this is considered as critical bug/feature, we can do the release now.

@cehbrecht @jorgejesus @tomkralidis  ?

tlvu:
We have 2 projects that would benefit from this fix right now.  It's because they have recently been rebuild and we found the issue with their new build.

But if we rebuild other projects for new release, then all the new builds will be affected by this bug as well.

Ideally, we would not want to make changes to all the various projects, then revert these changes later.

I am guessing other organizations consuming pywps are in the same situation.

I am totally for bundling new features in release but for regression bug like this one, could you make an exception and release right away?  Thanks.

huard:
Note that the current fix still passes the DefaultValue if it is None, which OWSLib does not handle. 

Not clear to me if this is desirable, since None is not a float per say. In other words, this fix does not entirely solve the original issue. I've seen similar issues with integers having a default set to inf. Should PyWPS enforce the rule that default values should be convertible to their type, or is this something left to the client ?

huard:
Sorry, my bad, I was confused. Everything looks ok.

jachym:
Ok, will do the release

jachym:
I hope, I managed it well https://pypi.org/project/pywps/4.2.3/

Issue #118, opened 2016-03-04
PYWPS_PROCESSES back


jachym:
reimplement process discovery as it was in pywps-3

tomkralidis:
:+1:

tomkralidis:
Related discussion: http://lists.osgeo.org/pipermail/pywps-dev/2016-March/000553.html

jachym:
During F2F meeting, Bonn, FOSS4G we agreed on:

* we will add configuration option to config file (no PYWPS_PROCESSES any more)
* we will create function to read the directory to pywps-core
* we have to make sure, once the files are changed, it will reload automatically (maybe documentation how to do it in apache etc. will be enough)
* @jorgejesus will investigate how to do it (keyworks: flask, docker, ....)

ldesousa:
This is also related to issue #322

cehbrecht:
See PR in pywps-flask: https://github.com/geopython/pywps-flask/pull/51

Issue #417, opened 2018-11-02
Dismiss operation (WPS 2.0.0)


huard:
# Description

Implement "dismiss" operation to stop an asynchronous process. 

Could someone split the tasks necessary to get there in small chunks that newbies can work on this ? 

For example: 
- [ ] Create dismiss XML template 
- [ ] Implement `cancel` method in `processing.basic.MultiProcessing`

cehbrecht:
@jachym Any comments on this issue?

cehbrecht:
See also #493

PR #505, opened 2019-12-16
initial watchdog

cehbrecht:
# Overview

This PR introduces a new `watchdog` module. The code includes the PR #497. 

Docs are not updated!

Changes:
* watchdog module to handle job queue.
* added click command line interface to start pywps and watchdog service.
* updated pywps config to load process list from python module.
* updated pywps configuration ... still some issues ... using `os.environ` `PYWPS_CFG`.

I have moved the CLI command line from the birdhouse template to pywps:
https://github.com/bird-house/cookiecutter-birdhouse/issues/52

You can start a pywps with the watchdog like this:
```
$ pywps start -c /path/to/pywps.cfg --use-watchdog
```

You can also start both services standalone:
```
$ pywps start -c /path/to/pywps.cfg
$ pywps watchdog -c /path/to/pywps.cfg
```

The `pywps.cfg` needs to point to the processes list with a python list, see Emu example:
https://github.com/bird-house/emu/blob/6347264a8fd0b86bf6d0b7b5fc857aaf449003cd/pywps.cfg#L10

This is a initial implemention to get the cli working. See also  issue #118.

# Related Issue / Discussion

#497 

# Additional Information

# Contribution Agreement


cehbrecht:
@jachym @davidcaron @huard could you please have a look at this PR and give your opinion. You can try it with the Emu on the `watchdog` branch:
https://github.com/bird-house/emu/tree/watchdog

cehbrecht:
Should the `watchdog` be started as its own command?
```
$ watchdog start -c pywps.cfg
```

cehbrecht:
We need also a `clean` task to remove stalled jobs from the `running` queue.

cehbrecht:
The scheduler `joblauncher` should be refactored to also load the job status from the database.

huard:
Started Emu with `pywps start --bind-host 0.0.0.0 -c pywps.cfg --use-watchdog`
then
```
from birdy import WPSClient
wps_a = WPSClient("http://localhost:5000/wps", progress=True)                                                                                                                                       
resp_a = wps_a.inout() 
``` 
fails with 
```
HTTPError: 500 Server Error: INTERNAL SERVER ERROR for url: http://localhost:5000/wps
```
No such problem in sync mode.

huard:
Also, would it be useful to have two logs, one for the daemon, and one for the server? Since the daemon generates a lot of noise in the logs (with DEBUG), it's hard to find the info relevant to the server.

cehbrecht:
> The daemon is sometimes called daemon, sometimes watchdog. I suggest a more descriptive name such as "RequestQueueDaemon".
> 

class name: `JobQueueHandler`?
module name: jobqueue or queue?

> There seems to be a mechanism in your PR to have a list of processes in the the configuration file. Is this documented ?

Not yet. It is a quick solution ... I needed this to get the CLI into pywps. See code:
https://github.com/cehbrecht/pywps/blob/7821562ef13c755b3b8ffab737a7c8241301308d/pywps/app/Service.py#L59

It is using the python expression of the process list and loads it via `importlib`.

huard:
+1

cehbrecht:
It is still working on my side ... I couldn't reproduce the exception.

> Started Emu with `pywps start --bind-host 0.0.0.0 -c pywps.cfg --use-watchdog`
> then
> 
> ```
> from birdy import WPSClient
> wps_a = WPSClient("http://localhost:5000/wps", progress=True)                                                                                                                                       
> resp_a = wps_a.inout() 
> ```
> 
> fails with
> 
> ```
> HTTPError: 500 Server Error: INTERNAL SERVER ERROR for url: http://localhost:5000/wps
> ```
> 
> No such problem in sync mode.

cehbrecht:
> Also, would it be useful to have two logs, one for the daemon, and one for the server? Since the daemon generates a lot of noise in the logs (with DEBUG), it's hard to find the info relevant to the server.

I'm not sure if I want to have two log files ... but I understand the trouble. As an alternative we could have two different types of logging handlers ... skip the db logs in debug mode ... unless they are wanted. Also the job queue can be triggered less frequently ... default was 30 secs.

cehbrecht:
@huard Do you have an opinion on the CLI interface?

huard:
Not really. Apart from watchdog name change, it looks good to me.

pywps/app/Service.py
```diff
@@ -58,6 +56,17 @@ def __init__(self, processes=[], cfgfiles=None):
             if not LOGGER.handlers:
                 LOGGER.addHandler(logging.NullHandler())
 
+        if not processes:
+            # load processes from processes_module.processes_list
+            pname = config.get_config_value('server', 'processes')
+            if pname:
+                pmodule = '.'.join(pname.split('.')[0:-1])
+                plist = pname.split('.')[-1]
```
davidcaron: `pmodule, plist = pname.rsplit('.', 1)`

pywps/cli.py
```diff
@@ -0,0 +1,116 @@
...
+            use_reloader=False,
+            threaded=True,
+            # processes=2,
+            use_evalex=True,
+            static_files=static_files)
+    # let's start the service ...
+    # See:
+    # * https://github.com/geopython/pywps-flask/blob/master/demo.py
+    # * http://werkzeug.pocoo.org/docs/0.14/serving/
+    if use_watchdog:
```
davidcaron: Did you think about starting the watchdog by default? Maybe that way getting started with pywps could be a bit simpler, I'm not sure...

pywps/configuration.py
```diff
@@ -135,20 +132,25 @@ def load_configuration(cfgfiles=None):
     CONFIG.add_section('grass')
     CONFIG.set('grass', 'gisbase', '')
 
+    CONFIG.add_section("watchdog")
+    CONFIG.set("watchdog", "pause", "30")
```
davidcaron: In the documentation, this section is called `daemon`.
Also, you could add the `pidfile` default configuration here?

pywps/dblog.py
```diff
@@ -170,13 +173,16 @@ def get_session():
                 engine = sqlalchemy.create_engine(database,
                                                   echo=echo,
                                                   connect_args={'check_same_thread': False},
-                                                  poolclass=NullPool)
```
davidcaron: Edit: The description below happens in the context of postgres, but I don't think it applies to sqlite... Even then,  the documentation for sqlalchemy mentions: 

> SQLite file-based connections have extremely low overhead, so pooling is not necessary. 

https://docs.sqlalchemy.org/en/13/dialects/sqlite.html#threading-pooling-behavior

I'm pretty sure we want to keep the NullPool here.

- The engine and the connection pool are a global variable. 
- The watchdog process has it in memory, opening connections to query if there are any jobs to do, and keeping these connections open because of this connection pooling.
- When forking happens, a copy of this connection pool is transferred to the forked process, with the same opened connections in the pool.
- When the worker is done processing, the connection pool is deleted and its connections are closed.
- When the watchdog process makes a query to check for any jobs to process, it will use a connection from its pool. If one of these connections had been previously closed, the watchdog will crash.

requirements.txt
```diff
@@ -1,4 +1,5 @@
-flufl.enum 
+click
+flufl.enum
```
davidcaron: can be removed if python 2.7 is not supported

docs/configuration.rst
```diff
@@ -195,8 +195,18 @@ configuration file <https://docs.pycsw.org/en/latest/configuration.html>`_.
   directory of the GRASS GIS instalation, refered as `GISBASE
   <https://grass.osgeo.org/grass73/manuals/variables.html>`_
 
+[daemon]
+--------
+
+:pidfile:
```
davidcaron: Not used?

pywps/app/Process.py
```diff
@@ -273,39 +261,11 @@ def _run_process(self, wps_request, wps_response):
             wps_response._update_status(WPS_STATUS.FAILED, msg, 100)
 
         finally:
```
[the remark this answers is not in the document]
cehbrecht: done.

pywps/app/Service.py
cehbrecht: done.

pywps/dblog.py
cehbrecht: done.

requirements.txt
cehbrecht: done.

pywps/app/Service.py
cehbrecht: done.

pywps/cli.py
cehbrecht: sure. But I also need to change the "watchdog" name.

cehbrecht: changed to:
```
pywps -c pywps.cfg --no-jobqueue
```

cehbrecht:
@huard @davidcaron 
renamed watchdog to:
```
class: pywps.queue.JobQueueService
cli: pywps jobqueue -c pywps.cfg
```

docs/configuration.rst
cehbrecht: updated docs. skipped `pidfile` in current implementation.

pywps/configuration.py
cehbrecht: updated to `jobqueue` in docs and code.

docs/deployment.rst
```diff
@@ -3,6 +3,47 @@
...
+
+-----------------
+`pywps_daemon.py`
+-----------------
+The :py:module:`bin.pywps_daemon` script is to be started separately - it will
+start daemon process, which will periodically check for the database stored
+requests and in case, some process is there, it will removed and executed.
+
+The daemon does start only so many processes, as many are allowed in the
+`maxparallel` process number.
```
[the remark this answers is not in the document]
cehbrecht: right. fixed.

docs/deployment.rst
```diff
@@ -3,6 +3,47 @@
 Deployment to a production server
 =================================
 
+PyWPS consists from 2 main parts: PyWPS :py:class:`pywps.app.Service` and the
+:py:module:`bin.pywps_daemon`
+
+
+The second part is :py:module:`bin.pywps_daemon` - the Daemon, which is
+running as background daemon process, responsible for calling of the stored
+request in the asynchronous mode.
```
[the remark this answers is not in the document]
cehbrecht: updated.

cehbrecht:
@huard I have introduced a new `pywps.log` module to remove code duplication, a config parameter `db_echo=false` to set database logging independently. The queue only logs in info mode when there is a change in the queue (running, stored). 

Probably logging can be improved in a later PR.

tlvu:
FYI I've built an emu docker image with tag `watchdog` here https://hub.docker.com/repository/docker/tlvu/emu/tags for those who wants to test using a real docker image.  I will try to bring this emu image into our PAVICS stack to see how it integrates.  Will report back any findings.

tlvu:
@cehbrecht do you know how provide Emu with extra configs?

I did this in my `docker-compose.yml` where I set  the env var `PYWPS_CFG=/wps.cfg`:
```
services:
  emu:
    image: ${EMU_IMAGE}
    container_name: emu
    environment:
      - PYWPS_CFG=/wps.cfg
    ports:
      - "8888:5000"
    volumes:
      - ./optional-components/emu/wps.cfg:/wps.cfg
      - wps_outputs:/data/wpsoutputs
      - /tmp
    restart: always
```

But it seems to ignore my `/wps.cfg` and only load the default one (`/opt/wps/pywps.cfg`):
```
$ docker exec emu cat /opt/wps/pywps.log
2019-12-18 22:56:11,703] [INFO] line=64 module=configuration loading configuration
2019-12-18 22:56:11,704] [INFO] line=151 module=configuration Configuration file(s) ['/opt/wps/pywps.cfg'] loaded
2019-12-18 22:56:12,008] [INFO] line=64 module=configuration loading configuration
2019-12-18 22:56:12,010] [INFO] line=151 module=configuration Configuration file(s) ['/opt/wps/pywps.cfg'] loaded
```

tlvu:
I also have another bird (Raven) on my system and the logs confirms it loads multiple config files, one of them is `/wps.cfg` which is added by the env var `PYWPS_CFG=/wps.cfg` the same way as Emu:

```
$ docker logs raven
INFO: configuration.load_configuration(): loading configuration
INFO: configuration.load_configuration(): Configuration file(s) ['/opt/wps/raven/default.cfg', '/opt/wps/.custom.cfg', '/opt/wps/etc/demo.cfg', '/wps.cfg'] loaded
starting WPS service on http://localhost:9099/wps
INFO: _internal._log():  * Running on http://0.0.0.0:9099/ (Press CTRL+C to quit)
```

cehbrecht:
@tlvu do you start the emu with:
```
pywps -c /wps.cfg
```

It will set the env `PYWPS_CFG=/wps.cfg` and it should be the only config used. Also the config needs to point to the processes list:
https://github.com/bird-house/emu/blob/2a6bac16433f22bad96241e9d1847b7572a8e776/pywps.cfg#L10

This this just an intermediate step to get things working.

tlvu:
@cehbrecht 
> @tlvu do you start the emu with:
> 
> ```
> pywps -c /wps.cfg
> ```
> 

I built the Emu docker image so the start command is the one from the Dockerfile 
https://github.com/bird-house/emu/blob/2a6bac16433f22bad96241e9d1847b7572a8e776/Dockerfile#L28


> It will set the env `PYWPS_CFG=/wps.cfg` and it should be the only config used. Also the config needs to point to the processes list:
> https://github.com/bird-house/emu/blob/2a6bac16433f22bad96241e9d1847b7572a8e776/pywps.cfg#L10
> 
> This is just an intermediate step to get things working.

Again, the default config is from the docker image so should be the one you referred to.

You can try my docker image here: `tlvu/emu:watchdog`.

This probably do not block me from testing the watchdog feature but I was simply very surprised that Emu seems to not load multiple configs anymore, like all the other birds.

Note I am taking a sick day today so won't be monitoring my messages.

cehbrecht:
@tlvu currently you can only provide a single config. It will be set by the `cli` module to `PYWPS_CFG`:
https://github.com/cehbrecht/pywps/blob/e6d4a2873a32ee0b0ee44b1703ff251452468a89/pywps/cli.py#L62
https://github.com/cehbrecht/pywps/blob/e6d4a2873a32ee0b0ee44b1703ff251452468a89/pywps/configuration.py#L146

This should be changed when fixing the configuration.

I tried the docker-compose in the emu and it was at least starting and responding. When building it don't forget the clean ... otherwise it will copy the existing database.
```
make clean
docker-compose build
docker-compose up
```

jachym:
> Should the `watchdog` be started as its own command?
> 
> ```
> $ watchdog start -c pywps.cfg
> ```

I think this does not matter. 

something like `pywps-deamon` would be maybe better, then neutral and not much informative `watchdog` ..

jachym:
I have created pull request for this pull request :-) https://github.com/cehbrecht/pywps/pull/1

cehbrecht:
@jachym @tlvu @davidcaron I have merged the alembic PR by Jachym and updated the command line to initalize the database. You need to run `migrate` before first start and on updates:
```
$ pywps migrate -c pywps.cfg
```

cehbrecht:
... ohh ... tests are failing ... no test DB.

cehbrecht:
@tlvu @jachym I have changed the command line ... `--config` option is now part of pywps command (not sub-command):
```
$ pywps -c pywps.cfg migrate
$ pywps -c pywps.cfg start
$ pywps -c pywps.cfg jobqueue
```

cehbrecht:
@jachym @davidcaron Should we change the config for the database settings in this PR or later?
See issue #495

cehbrecht:
@jachym @davidcaron We also need a "cancel" function for stalled jobs to clean the process queue. I already did some prototyping on it ... but I would prefer to get it in with a new PR.
See also issues #448 and #417.

jachym:
So, are we good? 
- [ ] Shall we merge?
- [ ] Release 4.4?

cehbrecht:
I would drop Python 2.7 before release 4.4. But that should be a quick change.

I'm ok with merging. Wait for feedback by: @tlvu  @davidcaron @elemoine (?)

> So, are we good?
> 
>     * [ ]  Shall we merge?
> 
>     * [ ]  Release 4.4?

alembic/README.md
```diff
@@ -0,0 +1,57 @@
...
+INFO  [alembic.runtime.migration] Context impl SQLiteImpl.
+INFO  [alembic.runtime.migration] Will assume non-transactional DDL.
+INFO  [alembic.runtime.migration] Running upgrade eead25e2ad3a -> 01d5298bfc61, Add process column to Request
+```
+
+## To make changes in database structure
+
+1. Change ORM
+
+        Currently, all changes in the datase are  done in `pywps/dblog.py` - make the
```
davidcaron: Currently, all changes in the database are done in `pywps/dblog.py` - make the
(typo in database and extra space)

davidcaron:
Sorry, I was still in holiday break.

@cehbrecht For the database config, you mean changing the default from sqlite in memory to a file-based sqlite? If so, I think this release is a good time to do it, whether in this PR or in another.

@jachym good catch for the database migration.

I have a suggestion also: considering that the command to start the service changes (new watchdog command to run jobs, migration is necessary on first start and on update) maybe it should be version 5 and not 4.4? There may not be 'enough' features to make a major version, but as a user of the library, I wouldn't expect breaking changes from 4.2 to 4.4 if we stick with [Semantic versioning](https://semver.org).

Apart from that, I haven't tested extensively, but the PR looks good to me.

cehbrecht:
> @cehbrecht For the database config, you mean changing the default from sqlite in memory to a file-based sqlite? If so, I think this release is a good time to do it, whether in this PR or in another.
> 

I had to check myself ... I meant to move the database config to a new config section, like mentioned in #495
> As a side note, IMO feel like logging.database is a bit misleading when this is not really specific to logging, probably should be server.database. My 2c anyway.

For the version number (4.4 or 5.0) I would leave it to @jachym. But this will be done after the merge.

tlvu:
I've deployed the emu test service on branch `watchdog` and it seems like the async mode is not working for me, see the following jupyterlab screenshot.  No problem in sync mode.

2020-01-08-162212_801x546_scrot

Matching logs:

```
$ docker exec emu tail /opt/wps/pywps.log
2020-01-08 21:24:21,498] [INFO] line=316 module=Service Request: getcapabilities
2020-01-08 21:24:22,025] [INFO] line=316 module=Service Request: describeprocess
2020-01-08 21:24:26,939] [INFO] line=316 module=Service Request: execute
2020-01-08 21:24:51,016] [INFO] line=37 module=queue PyWPS job queue: 0 running processes 1 stored requests
2020-01-08 21:24:51,031] [INFO] line=29 module=__init__ Processing mode: default
2020-01-08 21:24:51,123] [INFO] line=43 module=wps_say_hello say hello
2020-01-08 21:24:51,156] [INFO] line=266 module=Process Removing temporary working directory: /tmp/pywps_process_cfy81nlz
2020-01-08 21:25:21,086] [INFO] line=37 module=queue PyWPS job queue: 0 running processes 0 stored requests
```

You can access this notebook and reproduce this error here https://pavics.ouranos.ca/jupyter/user/public/lab/tree/users/lvu/test-pywps-watchdog/emu.ipynb.

The docker image is `tlvu/emu:watchdog-2257d76-20200108`.

Just for completeness, here is the working run in sync mode:

2020-01-08-163202_656x326_scrot

Matching logs:
```
$ docker exec emu tail -f /opt/wps/pywps.log
2020-01-08 21:31:30,589] [INFO] line=316 module=Service Request: getcapabilities
2020-01-08 21:31:31,006] [INFO] line=316 module=Service Request: describeprocess
2020-01-08 21:31:33,565] [INFO] line=316 module=Service Request: execute
2020-01-08 21:31:33,658] [INFO] line=43 module=wps_say_hello say hello
2020-01-08 21:31:33,685] [INFO] line=266 module=Process Removing temporary working directory: /tmp/pywps_process_ywvtght3
```

cehbrecht:
> I've deployed the emu test service on branch `watchdog` and it seems like the async mode is not working for me, see the following jupyterlab screenshot. No problem in sync mode.

@tlvu I can confirm. Looks like a python version issue. Seems to work in Python 3.6. In Python 3.8 I  had the following exception in the log file:
```
line=45 module=queue PyWPS job queue failed: Can't pickle <function do_first at 0x1061484c0>: it's not the same object as jinja2.filters.do_first
```

pywps/cli.py
```diff
@@ -0,0 +1,124 @@
+###########################################################
+# Demo WPS service for testing and debugging.
```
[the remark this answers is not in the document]
cehbrecht: done.

pywps/configuration.py
```diff
@@ -179,48 +177,16 @@ def checkdir(confid):
...
-                cfgfiles = (os.path.join(pywps.__path__[0], "etc", "pywps.cfg"), "/etc/pywps.cfg",
-                            os.path.join(os.getenv("HOME"), ".pywps.cfg"))
-            else:
-                cfgfiles = (os.path.join(pywps.__path__[0], "etc",
-                            "pywps.cfg"), "/etc/pywps.cfg")
-
+    LOGGER.debug('trying to estimate the default location')
+    cfgfiles = [os.path.join(pywps.__path__[0], "etc", "pywps.cfg"), "/etc/pywps.cfg"]
+    if 'HOME' in os.environ:
+        cfgfiles.append(os.path.join(os.environ['HOME'], ".pywps.cfg"))
```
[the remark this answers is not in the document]
cehbrecht: We already do not support windows ... so I cleaned the code to make it easier to read. `PYWPS_CFG` should overwrite all other configs (defaults, custom cfgfiles, ...)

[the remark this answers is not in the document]
tlvu: Oh was not aware we did not support Windows, thanks for clarification.

cehbrecht:
@tlvu The Python 3.8 failure is a different subject (see issue #508). For Python 3.6/3.7 it works ... but sometimes it does not. I have not yet figured out why. 


> > I've deployed the emu test service on branch `watchdog` and it seems like the async mode is not working for me, see the following jupyterlab screenshot. No problem in sync mode.
> 
> @tlvu I can confirm. Looks like a python version issue. Seems to work in Python 3.6. In Python 3.8 I had the following exception in the log file:
> 
> ```
> line=45 module=queue PyWPS job queue failed: Can't pickle <function do_first at 0x1061484c0>: it's not the same object as jinja2.filters.do_first
> ```

cehbrecht:
@jachym I have tried to update an existing pywps database (master) but this fails because the `pywps_stored_requests` table was already there:
```
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) table pywps_stored_requests already exists
```
Can we fix this in the upgrade script?

Also pywps has the possibility to configure a table prefix:
https://github.com/cehbrecht/pywps/blob/935197d2750a768788d9450a112ffa2bc5653a39/pywps/dblog.py#L53

Do we want to keep this? It will not be part of the upgrade script:
https://github.com/cehbrecht/pywps/blob/935197d2750a768788d9450a112ffa2bc5653a39/pywps/alembic/versions/eead25e2ad3a_create_tables.py#L34

cehbrecht:
@tlvu ... when ever I try to reproduce the failure everything works :( I thought something with the database migration went wrong. What happens if you start with a clean database?

> @tlvu The Python 3.8 failure is a different subject (see issue #508). For Python 3.6/3.7 it works ... but sometimes it does not. I have not yet figured out why.
> 
> > > I've deployed the emu test service on branch `watchdog` and it seems like the async mode is not working for me, see the following jupyterlab screenshot. No problem in sync mode.
> > 
> > 
> > @tlvu I can confirm. Looks like a python version issue. Seems to work in Python 3.6. In Python 3.8 I had the following exception in the log file:
> > ```
> > line=45 module=queue PyWPS job queue failed: Can't pickle <function do_first at 0x1061484c0>: it's not the same object as jinja2.filters.do_first
> > ```

tlvu:
> @tlvu ... when ever I try to reproduce the failure everything works :( I thought something with the database migration went wrong. What happens if you start with a clean database?

@cehbrecht emu on my instance (docker container) was starting from a clean database because the default DB is sqlite which is inside the container and not persisted outside the container.

I see you pin python 3.6 in emu.  I"ll get another docker build of emu and test that out later.

Note my broken emu image is `tlvu/emu:watchdog-2257d76-20200108` if you want to try re-producing inside a clean docker container without any DB migration.

pywps/cli.py
```diff
@@ -0,0 +1,124 @@
...
+@cli.command()
+@click.option('--bind-host', '-b', metavar='IP-ADDRESS', default='127.0.0.1',
+              help='IP address used to bind service.')
+@click.option('--no-jobqueue', '-n', is_flag=True,
+              help='Do not start job queue service.')
+def start(bind_host, no_jobqueue):
+    """Start PyWPS service.
+    This service is by default available at http://localhost:5000/wps
+    """
+    app = Service()
```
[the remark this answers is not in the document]
cehbrecht: made an update to allow overwrite of pywps config with env `PYWPS_CFG`:
https://github.com/geopython/pywps/pull/505/commits/1d0d3d86e193b848cb77465557ea40831f27740d

The migrate command is still using `PYWPS_CFG` and can not be overwritten.

It would possible to add multiple configs at the command line ... not sure if we want this.

cehbrecht:
@jachym I have the feeling we can move the pywps configuration in the alembic `env.py`:
https://github.com/cehbrecht/pywps/blob/1d0d3d86e193b848cb77465557ea40831f27740d/pywps/alembic/env.py#L13

... to the migrate command:
https://github.com/cehbrecht/pywps/blob/1d0d3d86e193b848cb77465557ea40831f27740d/pywps/cli.py#L123

Using `alembic.config.Config`:
https://alembic.sqlalchemy.org/en/latest/api/config.html

I need to pass the `pywps.cfg` files. Or I can use the alembic config just to pass `pywps.cfg` files and load them in alembic `env.py`.

alembic/README.md
cehbrecht: done.

cehbrecht:
> Note my broken emu image is tlvu/emu:watchdog-2257d76-20200108 if you want to try re-producing inside a clean docker container without any DB migration.

@tlvu I tried the image locally and it worked for me.

tlvu:
> @tlvu I tried the image locally and it worked for me.

@cehbrecht you mean the async worked?  That's odd.

Can you provided me with steps to reproduce on my side?

cehbrecht:
> > @tlvu I tried the image locally and it worked for me.
> 
> @cehbrecht you mean the async worked? That's odd.
> 
> Can you provided me with steps to reproduce on my side?

I didn't do anything special:
```
docker run -p 5000:5000 tlvu/emu:watchdog-2257d76-20200108 
```
Then I was starting several `hello` processes in async mode:
http://localhost:5000/wps?service=wps&version=1.0.0&request=execute&identifier=hello&datainputs=name=stranger&status=true&storeExecuteResponse=true

Checked the output of the last started process and it was ending successfully.

tlvu:
@cehbrecht 

I found it, the statusLocation was badly configured in my emu instance since I was not able to override it using `PYWPS_CFG`.

2020-01-13-114851_686x573_scrot

So I went into the container and manually change the config file and async is working

2020-01-13-122440_678x381_scrot

So I think this PR is good to go. I haven't tested yet your new commit that allow override using `PYWPS_CFG` but it will surely be immensely useful.

tlvu:
Oh by the way, how do I see the list of queued jobs? And is there a way to delete a queued job?

tlvu:
I've build `tlvu/emu:watchdog-0b36f1b-20200113` with python 3.6 pinned (don't need anymore) and pywps containing the new commit that allow override using `PYWPS_CFG`.

So my emu instance is behaving correctly against async request because the statusLocation can be properly set.

However in the logs, it does not seem to be reading multiple config files as expected (the default one, then the one specified by `PYWPS_CFG`) but only the one file specified by `PYWPS_CFG`.  Is this just a missing proper logs problem?
```
$ docker exec emu head  /opt/wps/pywps.log
2020-01-13 17:40:31,569] [INFO] line=64 module=configuration loading configuration
2020-01-13 17:40:31,570] [DEBUG] line=67 module=configuration setting default values
2020-01-13 17:40:31,570] [DEBUG] line=186 module=configuration trying to estimate the default location
2020-01-13 17:40:31,571] [INFO] line=151 module=configuration Configuration file(s) ['/wps.cfg'] loaded
2020-01-13 17:40:31,814] [DEBUG] line=27 module=literalvalidator validating literal value.
2020-01-13 17:40:31,814] [DEBUG] line=32 module=literalvalidator validation result: True
```

cehbrecht:
@jachym I have hard-coded the default database prefix:
https://github.com/geopython/pywps/pull/505/commits/2099b6e4494923eb9b4c5284b079519f295a4e47

Is this ok?

cehbrecht:
@tlvu I tried the configuration overwrite locally with emu. It worked for me and all the configs are written to the log file:
```
2020-01-14 14:09:39,986] [INFO] line=64 module=configuration loading configuration
2020-01-14 14:09:39,987] [INFO] line=151 module=configuration Configuration file(s) ['/Users/pingu/Documents/GitHub/birdhouse/emu/test.cfg'] loaded
2020-01-14 14:09:40,096] [INFO] line=64 module=configuration loading configuration
2020-01-14 14:09:40,097] [INFO] line=151 module=configuration Configuration file(s) ['pywps.cfg', '/Users/pingu/Documents/GitHub/birdhouse/emu/test.cfg'] loaded
2020-01-14 14:09:40,098] [INFO] line=64 module=configuration loading configuration
2020-01-14 14:09:40,104] [INFO] line=151 module=configuration Configuration file(s) ['pywps.cfg', '/Users/pingu/Documents/GitHub/birdhouse/emu/test.cfg'] loaded
```

There is still one config init step with the emu application:
https://github.com/bird-house/emu/blob/0b36f1baabaa33c5e16a60438e333d2a44c410b1/emu/__init__.py#L9

The app gets initialized without having the configs from the command line.

There were also config init steps with the command line config files in pywps. I have disabled them:
https://github.com/geopython/pywps/commit/2099b6e4494923eb9b4c5284b079519f295a4e47

cehbrecht:
@tlvu 
> Oh by the way, how do I see the list of queued jobs? And is there a way to delete a queued job?

I want to add the cancel function with a follow-up PR.

tlvu:
> @tlvu I tried the configuration overwrite locally with emu. It worked for me and all the configs are written to the log file:

@cehbrecht I think I understood why the emu docker image did not find any default config files:

https://github.com/cehbrecht/pywps/blob/2099b6e4494923eb9b4c5284b079519f295a4e47/pywps/configuration.py#L178-L183
        1. `PYWPS_INSTALLATION/etc/pywps.cfg`
        2. `/etc/pywps.cfg`
        3. `$HOME/.pywps.cfg`

https://github.com/bird-house/emu/blob/0b36f1baabaa33c5e16a60438e333d2a44c410b1/Dockerfile#L31
Default config file in emu docker image: `/opt/wps/pywps.cfg`

As we see, the default config file inside the emu docker image (/opt/wps/pywps.cfg) is not in any of the 3 locations above.  Should we modify emu docker image to put the default config file in 1 of those 3 locations or add more default locations?

cehbrecht:
> As we see, the default config file inside the emu docker image (/opt/wps/pywps.cfg) is not in any of the 3 locations above. Should we modify emu docker image to put the default config file in 1 of those 3 locations or add more default locations?

@tlvu we may want to add an additional default location for searching in the app path. Maybe we can do this in another PR? I think we have to change a bit more for this. Would it be ok to use `/etc/pywps.cfg` as an intermediate step? You only have on pywps in your container.

tlvu:
@cehbrecht 

> @tlvu we may want to add an additional default location for searching in the app path. Maybe we can do this in another PR? I think we have to change a bit more for this. 

Sure I am fine with this.

> Would it be ok to use `/etc/pywps.cfg` as an intermediate step? You only have on pywps in your container.

Yes for how we can drop the default config file in `/etc/pywps.cfg` instead of `/opt/wps/pywps.cfg`, since as you said, there's only 1 pywps per container so this is perfectly fine.

cehbrecht:
@jachym @tlvu @davidcaron What should we do with the alembic database upgrade? Currently the alembic scripts are not aware that there is an old db with a different schema and the migration step will fail. Should we drop the old DB manually? Or should we add the following to the alembic upgrade script (https://github.com/cehbrecht/pywps/blob/2099b6e4494923eb9b4c5284b079519f295a4e47/pywps/alembic/versions/eead25e2ad3a_create_tables.py#L19):

```
    # remove previous db
    try:
        op.drop_table('pywps_requests')
        op.drop_table('pywps_stored_requests')
    except Exception:
        pass
```

davidcaron:
A couple things to consider, and my opinions:
- Do we want to keep the history of `pywps_requests` or not? (I would personally not mind if it was dropped)
- The `pywps_`prefix in the table is configurable. (Alembic could have access to this configuration value, but I don't think the migration command parses the pywps configuration files.)
- If the schemas for `pywps_requests` and `pywps_jobs` at the first revision are the same, we could rename `pywps_requests` to `pywps_jobs` ... I think this would cover most of the environments if not all.
- Same thing for `pywps_stored_requests`, if it exists, keep it.

cehbrecht:
> A couple things to consider, and my opinions:
> 
>     * Do we want to keep the history of `pywps_requests` or not? (I would personally not mind if it was dropped)

I would prefer to drop it.

> 
>     * The `pywps_`prefix in the table is configurable. (Alembic could have access to this configuration value, but I don't think the migration command parses the pywps configuration files.)
> 

I have removed the configuration for the *prefix*. It is now fixed to `pywps_`.

>     * If the schemas for `pywps_requests` and `pywps_jobs` at the first revision are the same, we could rename `pywps_requests` to `pywps_jobs` ... I think this would cover most of the environments if not all.
> 
>     * Same thing for `pywps_stored_requests`, if it exists, keep it.

I would prefer to just drop the old tables. We just want to get in the alembic managed mode.

tlvu:
I suppose there's no real value to keep the existing data so we can drop those tables but I think @huard might understand better the implication of flushing existing data.

On the flip side, if the existing data is not so useful to preserve, why are we preserving them in the first place?

cehbrecht:
> On the flip side, if the existing data is not so useful to preserve, why are we preserving them in the first place?

we need a database for the job queue. PyWPS also uses the database for request logging ... but those infos are not meant to be kept for ever.

cehbrecht:
> @jachym @tlvu @davidcaron What should we do with the alembic database upgrade? Currently the alembic scripts are not aware that there is an old db with a different schema and the migration step will fail. Should we drop the old DB manually? Or should we add the following to the alembic upgrade script (https://github.com/cehbrecht/pywps/blob/2099b6e4494923eb9b4c5284b079519f295a4e47/pywps/alembic/versions/eead25e2ad3a_create_tables.py#L19):
> 
> ```
>     # remove previous db
>     try:
>         op.drop_table('pywps_requests')
>         op.drop_table('pywps_stored_requests')
>     except Exception:
>         pass
> ```

@jachym @tomkralidis @ldesousa Can you give me advise on this? I think this is the last thing to fix before we can merge.

tomkralidis:
I don't know the full context, but for data/database deletion I would prompt the user if they would like to drop the data in question, and if applicable add a `-y/--yes` flag if driven from a CLI.  So the underlying function that calls the table dropping has a `force=True/False` type flag which can be acted upon accordingly.

cehbrecht:
@tomkralidis I could extend the `pywps migrate` cli with  `--force` option to drop the previous database. I would leave the alembic untouched. At least this would help on the transition to alembic.

cehbrecht:
@tomkralidis I have added a force option to the migrate cli to drop the existing db: 
```
pywps migrate --force
```
https://github.com/geopython/pywps/pull/505/commits/c8ac7dd28c2aa9424eb92d2fafe7c685979c4779

jachym:
Everybody is happy ? Squash and merge? @tomkralidis ? @cehbrecht ? @huard and @davidcaron ? 

please add your review check

cehbrecht:
> Everybody is happy ? Squash and merge? @tomkralidis ? @cehbrecht ? @huard and @davidcaron ?
> 
> please add your review check

@jachym I'm happy with it. Should we make a 4.2.4 release before and/or open a 4.2.x branch?
https://github.com/geopython/pywps/compare/4.2.3...master

jachym:
> Should we make a 4.2.4 release before and/or open a 4.2.x branch?
> [4.2.3...master](https://github.com/geopython/pywps/compare/4.2.3...master)

I agree with making 4.2 branch and make relase

cehbrecht:
> > Should we make a 4.2.4 release before and/or open a 4.2.x branch?
> > [4.2.3...master](https://github.com/geopython/pywps/compare/4.2.3...master)
> 
> I agree with making 4.2 branch and make release.
> 
> IMHO since this is compability break, maybe the next release should be called 5.x ?

I'm ok with 5.x.

jachym:
merged to master  - thank you all

PR #513, opened 2020-02-07
Add DAP storage option

### LINKED ISSUE #512: opendap support for outputs
# Description
Our platforms runs a number of different pywps servers. The outputs of the process are stored on a disk that is indexed by THREDDS. THREDDS offers remote access to netCDF files using DAP. So what we'd like to be able to do is for processes to return an opendap url if the data_format is set to DODS. 

I think this can be done by 
- adding a `[dap]` configuration section indicating the host name;
- implementing a DapStorage, subclassing FileStorage and overriding the `url` method to return the dap endpoint indicated in the configuration.
- Return the DapStorage URL in `ComplexOutput._json_reference` if the data_format is DODS, otherwise return the FileStorage URL. 

Does this sound reasonable ?

huard:
# Overview

Fix #512 (in theory). Untested in practice, will need a DAP server running alongside a WPS server to test. 

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/inout/outputs.py
+++ b/pywps/inout/outputs.py
@@ -12,6 +12,7 @@ from pywps.app.Common import Metadata
 from pywps.exceptions import InvalidParameterValue
 from pywps.inout import basic
 from pywps.inout.storage.file import FileStorageBuilder
+from pywps.inout.storage.dap import DapStorageBuilder
 from pywps.validator.mode import MODE
 from pywps import configuration as config
 from pywps.inout.formats import Format
@@ -201,7 +202,10 @@ class ComplexOutput(basic.ComplexOutput):
         if self.prop == 'url':
             data["href"] = self.url
         elif self.prop is not None:
-            self.storage = FileStorageBuilder().build()
+            if self.data_format.mime_type == "application/x-ogc-dods":
+                self.storage = DapStorageBuilder().build()
+            else:
+                self.storage = FileStorageBuilder().build()
             data["href"] = self.get_url()
 
         return data
--- a/pywps/configuration.py
+++ b/pywps/configuration.py
@@ -142,6 +142,11 @@ def load_configuration(cfgfiles=None):
     CONFIG.set('s3', 'encrypt', 'false')
     CONFIG.set('s3', 'region', '')
 
+    CONFIG.add_section('dap')
+    CONFIG.set('dap', 'url', 'http://localhost/dap')
+    CONFIG.set('dap', 'outputurl', 'http://localhost/dap/dodsC/outputs')
+    CONFIG.set('dap', 'outputpath', outputpath)
+
     if not cfgfiles:
         cfgfiles = _get_default_config_files_location()
 
--- a/pywps/inout/storage/builder.py
+++ b/pywps/inout/storage/builder.py
@@ -5,11 +5,13 @@
 
 from .s3 import S3StorageBuilder
 from .file import FileStorageBuilder
+from .dap import DapStorageBuilder
 import pywps.configuration as wpsConfig
 
 STORAGE_MAP = {
     's3': S3StorageBuilder,
-    'file': FileStorageBuilder
+    'file': FileStorageBuilder,
+    'dap': DapStorageBuilder,
 }
 
 
--- a/pywps/inout/storage/dap.py
+++ b/pywps/inout/storage/dap.py
@@ -16,4 +16,3 @@ class DapStorageBuilder(StorageImplementationBuilder):
         file_path = config.get_config_value('dap', 'outputpath')
         base_url = config.get_config_value('dap', 'outputurl')
         return FileStorage(file_path, base_url)
-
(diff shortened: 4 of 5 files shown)


jachym:
@huard I think, the PR is great improvement - even I myself do not use the DAP storage feature at the moment, I love the interoperability possibilities, it offers.

Only thing I'm missing in your PR is update of documentation - you have to add section to configuration AND I would love to see a chapter in [extensions](https://pywps.readthedocs.io/en/latest/extensions.html) too.

Is it possible to add some tests as well?

huard:
Yes, I'll try to launch a dap server with emu to test it's really working.

Ideally, we'd have a general solution for this. Imagine for example that a process creates a shapefile and you'd like to serve it using wms and wcs. What should the wps response be ?

jachym:
Hi,

this is one of the dream features, which used to be in PyWPS, but had to go due to lack of resources.

Anyway, IMHO, this should be steered by mimetype - if the mimetype is image/tiff, it can be still link to wms or wcs, but the result should be the file. If the result is the service, then the mimetype should be application/vnd.ogc.wms_xml and the response should be the capabilities document IMHO.

pywps/inout/outputs.py
```diff
@@ -201,7 +202,10 @@ def _json_reference(self, data):
         if self.prop == 'url':
             data["href"] = self.url
         elif self.prop is not None:
-            self.storage = FileStorageBuilder().build()
+            if self.data_format.mime_type == "application/x-ogc-dods":
```
cehbrecht: This does not check the configured storagetype ... whenever I have a `dods` file it will use the `DapStorageBuilder` even when it is not configured.

cehbrecht:
@huard I think the `DapStorage` type behaves different then the existing ones. For the existing ones you only use the existing storage type: file or S3. With DAP it applies only for opendap links. Other files (txt, json, ...) will still be returned via HTTP. 

Is the `DapStorage` an extension of the `FileStorage` which decides on the mime type where to store and which output url to return?

huard:
yes, that was the idea. 
Should everything be served by dap? Or should it be configurable?

pywps/inout/outputs.py
huard: Indeed.

huard:
Thanks @cehbrecht for your comments. I realize it was a pretty naive implementation. 

I just realized that in the case where we want everything to be stored on a DAP server, we only have to modify the FileStorage configuration. I suggest to close this PR for now and see if we can do everything we need with the current FileStorage implementation and a tweaked configuration.

cehbrecht:
> yes, that was the idea.
> Should everything be served by dap? Or should it be configurable?

I don't think we want to serve everything by DAP. I'm not sure if OpenDAP (and WMS, WCS, ...) is a storage type. Maybe an output service type (which do not have yet)? The services may still use the same storage: file, S3, ... Depending on the mimetype and the configuration the output urls would be adapted to HTTP, OpenDAP, WMS, ...

@jachym what would you think?

huard:
Something like: 
```
CONFIG.add_section('service')
CONFIG.set("application/x-ogc-dods", "{host}/dap/dodsC/outputs")
```
and then the `storage.url` method uses this configuration to override the "file" url if needed ?

cehbrecht:
> Something like:
> 
> ```
> CONFIG.add_section('service')
> CONFIG.set("application/x-ogc-dods", "{host}/dap/dodsC/outputs")
> ```
> 
> and then the `storage.url` method uses this configuration to override the "file" url if needed ?

or maybe like this:
```
CONFIG.add_section('service')
CONFIG.set("dap_url", "{host}/dap/dodsC/outputs")
CONFIG.set("wms_url", "{host}/wms/outputs")
```
PyWPS may know the mimetypes.

huard:
@cehbrecht I think explicit mime-types offer some advantages. You could have two instances of a WMS server, one for version 1.0 and another for version 2.0.

cehbrecht:
> @cehbrecht I think explicit mime-types offer some advantages. You could have two instances of a WMS server, one for version 1.0 and another for version 2.0.

Ok. But I suppose we need to configure it differently ... not using the mimetype as config parameter.

ping @jachym.

huard:
Like the format name: 'GEOJSON, JSON, SHP, GML, METALINK, META4, KML, KMZ, GEOTIFF,'
                                 'WCS, WCS100, WCS110, WCS20, WFS, WFS100,'
                                 'WFS110, WFS20, WMS, WMS130, WMS110,'
                                 'WMS100, TEXT, DODS, NETCDF, LAZ, LAS, ZIP,'
                                 'XML'

cehbrecht:
> Like the format name: 'GEOJSON, JSON, SHP, GML, METALINK, META4, KML, KMZ, GEOTIFF,'
> 'WCS, WCS100, WCS110, WCS20, WFS, WFS100,'
> 'WFS110, WFS20, WMS, WMS130, WMS110,'
> 'WMS100, TEXT, DODS, NETCDF, LAZ, LAS, ZIP,'
> 'XML'

yes ... that could work.

huard:
So will close this one and open another PR when I come up with a more general solution.

PR #510, opened 2020-02-04
Add support for multiple languages

davidcaron:
# Overview

The 3 requests types are translated (GetCapabilities, DescribeProcess, Execute). 

I translated only the title and abstract attributes for processes, inputs and outputs. Should something else be translated?

See the changes to `docs/process.rst` for an example of the interface.

# Contribution Agreement

- [x] I'd like to contribute this feature to PyWPS. I confirm that my contributions to PyWPS will be compatible with the PyWPS license guidelines at the time of contribution.
- [x] I have already previously agreed to the PyWPS Contributions and Licensing Guidelines

PR #519, opened 2020-02-12
fix tests with language option

cehbrecht:
# Overview

This PR fixes the current test failures with language option.

# Related Issue / Discussion

See PR #510.

# Additional Information

# Contribution Agreement

### DIFF
--- a/tests/test-pywps.cfg
+++ b/tests/test-pywps.cfg
@@ -1,3 +1,6 @@
+[server]
+language=en-US,fr-CA
+
 [logging]
 level=INFO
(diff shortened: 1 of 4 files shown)


cehbrecht:
@davidcaron the `CONFIG.set` overwrite of the config does not work in the tests anymore when using `pywps.tests.client_for`. To get the tests working again I have added the `language=en-US,fr-CA` to the test config.

cehbrecht:
@davidcaron I'm merging this to get the test suite running again. Maybe we can fix the config issue with another PR.

PR #522, opened 2020-02-13
Customize URL as a function of the data format

### LINKED ISSUE #512: opendap support for outputs

huard:
# Overview
This PR proposes a mechanism that customizes the URL of an output based on its data format. By default the standard outputurl is used, but if the data format (identified by the field of the FORMATS tuple) is included in the `services` section of the configuration, then this is substituted. 

# Related Issue / Discussion
Fixes #512 

The use case for this is to run a THREDDS server or GeoServer alongside a PyWPS server, and serve some of the process outputs using dedicated services instead of a simple file url. 

# Additional Information

# Contribution Agreement

### DIFF
--- a/default-sample.cfg
+++ b/default-sample.cfg
@@ -53,3 +53,6 @@ region=us-east-1
 prefix=appname/coolapp/
 public=true
 encrypt=false
+
+
+[services]
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -22,6 +22,7 @@ The configuration file has several sections:
     * `grass` for *optional* configuration to support `GRASS GIS
       <https://grass.osgeo.org>`_
     * `s3` for *optional* configuration to support AWS S3 storage
+    * `services` for custom output URLs based on output's mime-type
 
 PyWPS ships with a sample configuration file (``default-sample.cfg``).
 A similar file is also available in the `flask` service as
@@ -96,7 +97,7 @@ configuration file <https://docs.pycsw.org/en/latest/configuration.html>`_.
     the URL of the WPS service endpoint
 
 :language:
-    a comma-separated list of ISO 639-1 language and ISO 3166-1 alpha2 country 
+    a comma-separated list of ISO 639-1 language and ISO 3166-1 alpha2 country
     code of the service
     (e.g. ``en-CA``, ``fr-CA``, ``en-US``)
 
@@ -210,7 +211,7 @@ configuration file <https://docs.pycsw.org/en/latest/configuration.html>`_.
   <https://grass.osgeo.org/grass73/manuals/variables.html>`_
 
 [jobqueue]
---------
+----------
 
 :pause:
   pausing in seconds between periodical check for new stored requests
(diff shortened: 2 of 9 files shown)


huard:
If this sounds about right, I'll update the documentation.

cehbrecht:
With this PR we can configure different output urls depending on the mime-type (like WMS, OpenDAP, ...). The process does not need to worry about this. It uses the the pywps FORMATS for configuring the output url for a specific mime-type (like `WMS130`):
https://github.com/geopython/pywps/blob/d05483d75e753b3cda303f5c0bb778a0f9465393/pywps/inout/formats/__init__.py#L18

@tomkralidis @jachym @ldesousa Are you happy with this approach?

jachym:
This looks cool, but could I ask you for adding some more descriptive example in the documatation, where you could describe the integration with GeoServer? E.g. in https://pywps.readthedocs.io/en/latest/external-tools.html ?

huard:
I don't really know how it would work, because the url for WMS/WFS is not just a path, but a request. I'll need to help on this one.

huard:
From what I can quickly understand, adding a layer to GeoServer requires a call to the REST API, telling it in which workspace it should go, what is the default style, etc. The solution I propose in this PR is too basic for this. Saving to GeoServer would probably be done explicitly within the process handler: 
```
url = store_geoserver(feature, workspace, **kwargs)
request.response['output'].url = url
```

jachym:
ok, we have to work on the `store_geoserver` function, whatever it may be

from I have no more questions, we can merge

cehbrecht:
@huard ok to merge?

huard:
I'm concerned we will need to change the config to support the GeoServer use case. Do you want to mention in the docs that this feature should be considered experimental ? 

To support @jachym use case with GeoServer, I can see two options: 

1. Have a `store_geoserver` function called explicitly inside the process handler. Developer would be responsible for the configuration. 
2. Have a `GeoServerStorage` class, that is used for certain output mime-types.

In the case of 2, the configuration would probably be like: 

```
[storage]
DODS=thredds
WMS=geoserver
*=file [or s3]

[file]
host=
outputpath=

[geoserver]
host=
workspace=

[thredds]
host=
outputpath=
```

cehbrecht:
I think I would prefer option 2 ...having a `GeoServerStorage` class. Could this be done in another PR?

huard:
Yes, but then this PR should not be merged, as it will conflict.

jachym:
I suggest, we leave the geoserver usace for now (if you are not desperately looking forward to implement it) and and make some ticket.

The GeoServer should use `gsutil` python module. I think that this truly leads to GeoServer storage class - because "all the data" would be then distributed via geoserver serivices

And in the future, MapServer would be nice too ;-)

PR #523, opened 2020-02-24
allow custom UOM reference for literal IO

davidcaron:
# Overview

Currently, the only UOMs allowed when serializing an input are `degree`, `meter` and `unity` (see `pywps.OGCUNIT`). With the new 'watchdog' process, the inputs are always serialized, so these units are the only ones allowed.

I think that leaving the responsibility to the PYWPS service implementation to provide the UOM reference could be a simple solution to this.

# Contribution Agreement

### DIFF
--- a/pywps/inout/basic.py
+++ b/pywps/inout/basic.py
@@ -65,12 +65,16 @@ class UOM(object):
     :param uom: unit of measure
     """
 
-    def __init__(self, uom=''):
+    def __init__(self, uom='', reference=None):
         self.uom = uom
+        self.reference = reference
+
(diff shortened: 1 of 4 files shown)


cehbrecht:
@jachym This PR just extends the current UOMs implementation. We could add more common UOMs to the OGCUNIT list:
 https://github.com/geopython/pywps/blob/7126011c87d305dce22780b8611a3297a92b8a7a/pywps/__init__.py#L80

cehbrecht:
@davidcaron merged. Thanks for this contribution :)

PR #532, opened 2020-05-14
Issue 531 configparser with encoding

### LINKED ISSUE #531: Assume config-files are encoded in utf-8
In python 3.2 configparser added support for the encoding-parameter:
https://docs.python.org/3.5/library/configparser.html#configparser.ConfigParser.read
Before this configparser just opened the file in whatever open() expected, which again used whatever the local default encoding/character set was.

I'd suggest taking advantage of this update and changing the line which reads the file in pywps/configuration.py to the following: 
`loaded_files = CONFIG.read(config_files, encoding='utf-8')`

Right now pywps may or may not crash when reading the config-file depending on whether or not you're:
-using any non-ascii characters in the config file.
-what the default character set is for your specific machine.

I think it would've been better to always expect a utf-8 config-file, rather than letting it depend on the specific machine. 

This might not be viable if pywps still wants to support python 2.7 or python < 3.2

# Environment

- operating system: Centos 7
- Python version: 3.6
- PyWPS version: 4.2.4
- source/distribution
 - [X ] PyPI
- web server
 - [X ] Apache/mod_wsgi

# Steps to Reproduce
Create and load a config-file encoded as utf-8 containing one or more non-ascii characters. OBS: must be done on a machine where the non-ascii character is not a part of the character set used by open() .

# Additional Information

cehbrecht:
# Overview

* Added encoding to configparser as suggested by issue #531.
* Added supported python versions to `setup.cfg` (3.6, 3.7).

# Related Issue / Discussion

Fix #531 

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/configuration.py
+++ b/pywps/configuration.py
@@ -146,7 +146,7 @@ def load_configuration(cfgfiles=None):
     if 'PYWPS_CFG' in os.environ:
         config_files.append(os.environ['PYWPS_CFG'])
 
-    loaded_files = CONFIG.read(config_files)
+    loaded_files = CONFIG.read(config_files, encoding='utf-8')
     if loaded_files:
         LOGGER.info('Configuration file(s) {} loaded'.format(loaded_files))
     else:
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,9 @@ CONFIG = {
         'License :: OSI Approved :: MIT License',
         'Operating System :: OS Independent',
         'Programming Language :: Python',
+        "Programming Language :: Python :: 3",
+        "Programming Language :: Python :: 3.6",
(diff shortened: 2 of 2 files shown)


cehbrecht:
ping @kvold. Could also be added to the 4.2.x release.

cehbrecht:
travis py36 works ... but fails with netcdf issue for py37 ... not related to this PR.

kvold:
It would be great if this could be added to an 4.2.x release. The only possible issue I can see is if someone was relying on the dynamic behavior of open() to read a non-utf8 and non-ascii config file.

cehbrecht:
Added this PR to the 4.2.x branch ... so it should be part of the next 4.2.5 releaase.

PR #539, opened 2020-07-02
Fix scheduler

cehbrecht:
# Overview

This PR is a quick fix for scheduler extension. It make sure that `process.status_store` is not None.

It also fixes the `cfgfiles` in `configuration.py` ... a bug introduced by PR #532.

# Related Issue / Discussion

#532 

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/configuration.py
+++ b/pywps/configuration.py
@@ -149,9 +149,9 @@ def load_configuration(cfgfiles=None):
         cfgfiles = [cfgfiles]
 
     if 'PYWPS_CFG' in os.environ:
-        config_files.append(os.environ['PYWPS_CFG'])
+        cfgfiles.append(os.environ['PYWPS_CFG'])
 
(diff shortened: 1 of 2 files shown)


pywps/configuration.py
```diff
@@ -149,9 +149,9 @@ def load_configuration(cfgfiles=None):
         cfgfiles = [cfgfiles]
 
     if 'PYWPS_CFG' in os.environ:
-        config_files.append(os.environ['PYWPS_CFG'])
+        cfgfiles.append(os.environ['PYWPS_CFG'])
```
tlvu, line 152: Oh wow, good catch ! Without this fix, the default config file was completely ignored if `PYWPS_CFG` is set?

cehbrecht: The service did not start when `PYWPS_CFG` was set.

PR #542, opened 2020-07-28
ext_autodoc: support RST anonymous link

tlvu:
# Overview

Sphinx "Duplicate explicit target name" warning is problematic when
warning are turned into errors (`sphinx-build -W`).

And we would want warnings to turn into error because autodoc import
failure are treated as warnings only, resulting into silent
documentation build failure if those warnings do not fail the build.

# Related Issue / Discussion

See PR https://github.com/bird-house/cookiecutter-birdhouse/pull/96 with the silent doc build error.

See commit https://github.com/bird-house/flyingpigeon/pull/336/commits/c97ebca73792c1fb0c32d7b6079385fcd84a73b0 for actual real usage of this change.

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/ext_autodoc.py
+++ b/pywps/ext_autodoc.py
@@ -8,6 +8,21 @@ from pywps import Process
 from pywps.app.Common import Metadata
 
 
+class MetadataUrl(Metadata):
+    """Metadata subclass to allow anonymous links generation.
+
+    Useful to avoid Sphinx "Duplicate explicit target name" warning.
+
+    See https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#anonymous-hyperlinks.
+    """
+
+    def __init__(self, title, href=None, role=None, type_='simple',
+                 anonymous=False):
+        super().__init__(title, href=href, role=role, type_=type_)
+        self.anonymous = anonymous
+        "Whether to create anonymous link (boolean)."
+
+
 class ProcessDocumenter(ClassDocumenter):
     """Sphinx autodoc ClassDocumenter subclass that understands the
     pywps.Process class.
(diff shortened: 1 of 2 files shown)


tlvu:
This change is used in https://github.com/bird-house/flyingpigeon/pull/336

tlvu:
Ping @cehbrecht @huard

pywps/ext_autodoc.py
```diff
@@ -8,6 +8,21 @@
 from pywps.app.Common import Metadata
 
 
+class MetadataUrl(Metadata):
```
cehbrecht, line 11: I would add the `anonymous` option to the `Metadata` class itself instead of sub-classing it. What do you think?

tlvu: That was my initial thought until I saw how the `Metadata` class is being used.  It's being used to generate the xml response, it can be serialized to and deserialized from json format.  So bigger impact and not required for what I needed to solve.

So I am unsure if adding the new `anonymous` option there is appropriate.

But it's my first pywps commit so I will follow you guidance.  Let me know if adding it to the `Metadata` is the right thing to do.

cehbrecht: @tomkralidis What is your opinion?

tlvu:
Note to the person merging this PR, please do not use squash merge.

Squash merge will not merge the original commit and I have referenced the original commit in https://github.com/bird-house/flyingpigeon/blob/baf953244dee562b04b966dfb309987ed0ab588d/Makefile#L57 and in https://github.com/bird-house/emu/blob/3620ae99c5c5edb735e8b53e3bd0c1f0c650dd23/Makefile#L75

Alternatively please give me write access to https://github.com/geopython/pywps and I'll merge this PR myself, after approval or other required change of course.

cehbrecht:
> Note to the person merging this PR, please do not use squash merge.
> 
> Squash merge will not merge the original commit and I have referenced the original commit in https://github.com/bird-house/flyingpigeon/blob/baf953244dee562b04b966dfb309987ed0ab588d/Makefile#L57 and in https://github.com/bird-house/emu/blob/3620ae99c5c5edb735e8b53e3bd0c1f0c650dd23/Makefile#L75
> 
> Alternatively please give me write access to https://github.com/geopython/pywps and I'll merge this PR myself, after approval or other required change of course.

@tlvu I have given you write access as a contributor. Lets wait for feedback by @tomkralidis and @jachym.

cehbrecht:
@tlvu feel free to merge :) I can make a patch release afterwards.

Issue #553, opened 2020-12-08
How to handle json dump for inputs with CDATA?


cehbrecht:
# Description

When dumping (as json) a complex-input with format json then the data gets tagged with `CDATA`:
https://github.com/geopython/pywps/blob/9fa56cc17bf545a9a583271d83affa5b479d3373/pywps/inout/inputs.py#L243

Is this necessary?

This needs to be handled when loading the json dump again ... currently this is not the case.

See also #444.

# Environment

- operating system:
- Python version:
- PyWPS version: 4.2
- source/distribution
 - [ ] git clone
 - [ ] Debian
 - [ ] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [ ] Apache/mod_wsgi
 - [ ] CGI
 - [ ] other (please specify): 

# Steps to Reproduce

We have a workflow process with a workflow document in json:
https://github.com/roocs/rook/blob/858130631bf0a37c19a78e8e94961b7159846833/rook/processes/wps_orchestrate.py#L13

The json document is send with the wps request ... not as reference.

When we use the pywps scheduler extension the `WPSRequest` is dumped as a json file:
https://github.com/geopython/pywps/blob/9fa56cc17bf545a9a583271d83affa5b479d3373/pywps/processing/job.py#L71

In this case the workflow document (json format) is tagged by `CDATA`:
https://github.com/geopython/pywps/blob/9fa56cc17bf545a9a583271d83affa5b479d3373/pywps/inout/inputs.py#L243

When the json dump is loaded the `CDATA` tag is still part of the workflow document:
https://github.com/geopython/pywps/blob/9fa56cc17bf545a9a583271d83affa5b479d3373/pywps/processing/job.py#L81

... and the json loader for the workflow document will fail.
https://github.com/roocs/rook/blob/858130631bf0a37c19a78e8e94961b7159846833/rook/processes/wps_orchestrate.py#L52

# Additional Information

PR #444

cehbrecht:
@huard how would you handle this? 

Should we add the json data as it is without `CDATA`?
https://github.com/geopython/pywps/blob/9fa56cc17bf545a9a583271d83affa5b479d3373/pywps/inout/inputs.py#L227

Or when loading the json dump should the `CDATA` tag be removed?
https://github.com/geopython/pywps/blob/9fa56cc17bf545a9a583271d83affa5b479d3373/pywps/inout/inputs.py#L217

huard:
[...]

PR #555, opened 2020-12-10
fix cdata deserialzation

cehbrecht:
# Overview

This PR fixes issue #553. It removes the `CDATA` tag when loading an input from json.

# Related Issue / Discussion

#553 

# Additional Information

https://github.com/roocs/rook/pull/72

# Contribution Agreement

### DIFF
--- a/pywps/inout/inputs.py
+++ b/pywps/inout/inputs.py
@@ -214,7 +214,11 @@ class ComplexInput(basic.ComplexInput):
         elif json_input.get('href'):
             instance.url = json_input['href']
         elif json_input.get('data'):
-            instance.data = json_input['data']
+            data = json_input['data']
+            # remove cdata tag if it exists (issue #553)
+            if 'CDATA' in data:
+                data = data.replace("<![CDATA[", "").replace("]]>", "")
+            instance.data = data
 
         return instance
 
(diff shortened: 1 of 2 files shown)


pywps/inout/inputs.py
```diff
@@ -214,7 +214,11 @@ def from_json(cls, json_input):
         elif json_input.get('href'):
             instance.url = json_input['href']
         elif json_input.get('data'):
-            instance.data = json_input['data']
+            data = json_input['data']
+            # remove cdata tag if it exists (issue #553)
+            if 'CDATA' in data:
+                data = data.replace("<![CDATA[", "").replace("]]>", "")
```
huard, line 220: I think it would be safer to use a regex here. If the data contains instances of "]]>", this will have unintended consequences.

huard (approved):
Note that another, maybe cleaner option, would be to set the CDATA tag into the xml template, instead of the json representation. 

I might have tried to do that originally but could not get all the kinks out. It might be more difficult though to code the necessary logic inside the template.

pywps/inout/inputs.py
cehbrecht: I have added a regex for cdata.

cehbrecht:
> Note that another, maybe cleaner option, would be to set the CDATA tag into the xml template, instead of the json representation.
> 
> I might have tried to do that originally but could not get all the kinks out. It might be more difficult though to code the necessary logic inside the template.

I have opened an issue for this #557.

PR #563, opened 2020-12-17
fix output stream of scheduler

cehbrecht:
# Overview

This PR splits the output stream for `error` and `stdout` in different files. This is necessary on some parallel filesystems.

# Related Issue / Discussion

# Additional Information

# Contribution Agreement


cehbrecht:
@tomkralidis merging it has a quick-fix.

Issue #568, opened 2021-01-27
Long pending changes of python version compatibilities


fmigneault:
# Description

PR https://github.com/geopython/pywps/pull/520 seems to indicate that many fixes where made to resolve dropping Python 2 and errors during handling of mismatching dependencies across python versions. 

As of yesterday, version `4.2.10` still does not include them. They are only 'pending' in master: 
https://github.com/geopython/pywps/compare/master..4.2.10

Every tag since `4.2.4` are in parallel and diverging from master, without any backported fixes. 

Since pip 21.0 enforces stronger dependency validations with their new resolver, it now becomes impossible to "cleanly" install `pywps`. 
Either `pathlib` is missing and `pywps` complains, or it is installed and many other packages break due to overridden builtin `pathlib`.

A proper sync between master / release tags should be done once and for all. 
Why are there even releases of features not integrated in master?

fmigneault:
@cehbrecht How should this be handled?

cehbrecht:
@fmigneault sorry about the mess. We have merged several things to quickly to master (dropping py2 and proposed job queue). The new job queue implementation on master is not used and probably we want a different implementation.

To get out of this I would propose the following:
* make a pywps 4.4 branch for python 3 only
* we need to create this branch from master with the already python 3 only PRs (#520).
* we merge the PRs from 4.2 branch into 4.4 branch.

Maybe we can move the current master to a dev branch for the current job queue implementation.
The pywps 4.4 branch would then become the new master.

After that we need to think again about finding a solution for the job queue implementation.

cehbrecht:
I could do the following:
* create a new pywps-4.4 branch based on pywps-4.2
* backport (cherry-pick) patches from master (excluding new job queue)
* clean up: py3 only, test
* make a 4.4.0 release

Next:
* move current master to a dev branch
* make the pywps-4.4 branch the new master

@jachym @tomkralidis @huard Thoughts?

huard:
Sounds good !

tomkralidis:
+1

fmigneault:
> Maybe we can move the current master to a dev branch for the current job queue implementation.
The pywps 4.4 branch would then become the new master.

I like this idea. Experimental features should be in a separate branch and merged only when deemed appropriate. 
The most confusing thing in this repo is that its landing page defaults to master which is not aligned with recent releases. 

The plan sounds good. 
I'm curious, why it jumps from pywps-4.2 to pywps-4.4 ?

[...]

PR #569, opened 2021-01-27
drop py2 reqs

fmigneault:
# Overview

The bigger issue about desync master / pywps-4,2 branch remains, but at least pywps-4.2 branch can be installed without conflicting dependencies. 

# Related Issue / Discussion
#568 

# Additional Information

# Contribution Agreement

- [x] I'd like to contribute #568 to PyWPS. I confirm that my contributions to PyWPS will be compatible with the PyWPS license guidelines at the time of contribution.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,5 @@
 coverage
 flake8
-flufl.enum
 pylint
 six
 Sphinx
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,10 @@
-flufl.enum 
+flufl.enum; python_version < "3"
+enum34; python_version == "3.4"
 jinja2
 jsonschema
 lxml
 owslib
-pathlib
+pathlib; python_version < "3.4"
 python-dateutil
 requests
 SQLAlchemy
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,10 +1,7 @@
-flufl.enum; python_version < "3"
-enum34; python_version == "3.4"
 jinja2
 jsonschema
 lxml
 owslib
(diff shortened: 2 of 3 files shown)


cehbrecht:
I would prefer to support only `python>=3.6`. 

`flufl.enum` has support for python 3.6 ... well ... by doing nothing.
https://pypi.org/project/flufl.enum/4.1.1/

Can we skip `enum34` and `pathlib` if we only support `py>=3.6`?

Dropping complete py2 support would go into the pywps 4.4 version.

cehbrecht (commented):
see inline

fmigneault:
I don't mind dropping their support. I'm also good with >=3.6 has this is the versions I use. 
The critical issue I face is that those libraries somehow get installed when they shouldn't. 
They end up overriding the ones in standard libs, which result in missing methods being raised.

cehbrecht:
> I don't mind dropping their support. I'm also good with >=3.6 has this is the versions I use.
> The critical issue I face is that those libraries somehow get installed when they shouldn't.
> They end up overriding the ones in standard libs, which result in missing methods being raised.

I would only support Python >=3.6. Would you like to make this change so that it fits?

cehbrecht (approved):
see inline

fmigneault:
@cehbrecht all done. 
I also added the `python_requires` for setup install. This will be reflected in pip packages/releases.

cehbrecht (approved):

cehbrecht:
@fmigneault looks good :smile: I will make a new 4.2 release ... hopefully the last one. The next one should be 4.4.

PR #578, opened 2021-02-11
fix incorrect use of self.__class__ in super

fmigneault:
# Overview

Fix incorrect use of `self.__class__` in `super` call.

# Additional Information

In the actual context, it just so happens that `self.__class__` is equal to `ExecuteResponse` and produces the same result. 
When that class is derived though, to add additional functionalities, the value changes and causes endless recursion. 

```
class DerivedExecuteResponse(ExecuteResponse): 
    def __init__(self, wps_request, uuid):
        super(ExecuteResponse, self).__init__(wps_request, uuid)
        # other setup... 

class ExecuteResponse(WPSResponse):
    def __init__(self, wps_request, uuid, **kwargs):
        # here 'self.__class__'  == 'DerivedExecuteResponse'
        # 'super(self.__class__, self)' resolves to 'ExecuteResponse', so it calls itself !
        super(self.__class__, self).__init__(wps_request, uuid)
```

# Contribution Agreement

### DIFF
--- a/pywps/response/execute.py
+++ b/pywps/response/execute.py
@@ -33,7 +33,7 @@ class ExecuteResponse(WPSResponse):
         :param uuid: string this request uuid
         """
 
-        super(self.__class__, self).__init__(wps_request, uuid)
+        super(ExecuteResponse, self).__init__(wps_request, uuid)
 
         self.process = kwargs["process"]
         self.outputs = {o.identifier: o for o in self.process.outputs}


fmigneault:
@cehbrecht 
Would be great if that can be integrated in #577 
So far only found this problem while using 4.4, the rest seems to work correctly.

cehbrecht (approved):

cehbrecht:
@fmigneault Thanks for the fix 👍 I will update the release notes.

PR #579, opened 2021-02-18
Fix issue with stored request

gschwind:
Stored request was not restored correctly in the function
Process.launch_next_process, in particular "asReference" outputs aren't
"asReference" any more after being restored.

# Overview

# Related Issue / Discussion

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/app/Process.py
+++ b/pywps/app/Process.py
@@ -21,9 +21,11 @@ from pywps.inout.outputs import output_from_json
 import pywps.configuration as config
 from pywps._compat import PY2
 from pywps.exceptions import (StorageNotSupported, OperationNotSupported,
-                              ServerBusy, NoApplicableCode)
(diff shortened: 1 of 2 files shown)


pywps/app/Process.py
```diff
@@ -313,6 +315,7 @@ def launch_next_process(self):
             process._set_uuid(uuid)
             process._setup_status_storage()
             process.async_ = True
+            process.setup_outputs_from_wps_request(new_wps_request)
```
cehbrecht, line 318: This line is the main change in the PR. The rest is just creating a method for this code.

cehbrecht:
I would have expected the process would have been completed restored from the json dump.

cehbrecht (approved):

cehbrecht:
@gschwind merged. thanks :)

PR #580, opened 2021-03-17
fixed file size check for outputs

### LINKED ISSUE #571: Unit mismatch in FileSizeExceeded error_message
# Description

The `file` property of the `URLHandler` object checks if the data size is larger than the maximum data size. When it exceeds the maximum value, it raises an error: 
https://github.com/geopython/pywps/blob/20e1e254a3f7914e555fa89f363d1f6eb5f3895c/pywps/inout/basic.py#L458

The value is in units of bytes, but the message says megabytes. 

Also, it would be useful if the message included the actual size of the data. 

# Environment

- operating system:
- Python version: 3.8
- PyWPS version:4.2.9
- source/distribution
 - [x] git clone
 - [ ] Debian
 - [ ] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [ ] Apache/mod_wsgi
 - [ ] CGI
 - [ ] other (please specify): 

# Steps to Reproduce

# Additional Information

cehbrecht:
# Overview

This PR fixes the file size check for outputs.

The creation of an output with metalink failed when the output file exceeded the `maxsingleinputsize` value.
* updated docs
* fixed exception catch of `FileSizeExceeded`
* using https://pypi.org/project/humanize/

The root cause for this error is fixed in the PR #581.

# Related Issue / Discussion

Fix #571 

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/inout/basic.py
+++ b/pywps/inout/basic.py
@@ -517,7 +517,8 @@ class UrlHandler(FileHandler):
 
         return req
 
-    def max_size(self):
+    @staticmethod
+    def max_size():
         """Calculates maximal size for input file based on configuration
         and units.
 
--- a/pywps/inout/basic.py
+++ b/pywps/inout/basic.py
@@ -521,8 +521,7 @@ class UrlHandler(FileHandler):
         :return: maximum file size in bytes
         """
         ms = config.get_config_value('server', 'maxsingleinputsize')
-        mb_size = config.get_size_mb(ms)
-        byte_size = mb_size * 1024**2
+        byte_size = config.get_size_mb(ms) * 1024**2
         return byte_size
 
 
@@ -1030,14 +1029,3 @@ class ComplexOutput(BasicIO, BasicComplex, IOHandler):
         (_, _, url) = self.storage.store(self)
         # url = self.storage.url(self)
         return url
-
-    def max_size(self):
-        """Calculates maximal size for inout file based on configuration
-        and units.
-
-        :return: maximum file size in bytes and megabytes
-        """
-        ms = config.get_config_value('server', 'maxsingleinputsize')
-        mb_size = config.get_size_mb(ms)
-        byte_size = mb_size * 1024**2
-        return byte_size, mb_size
--- a/pywps/inout/basic.py
+++ b/pywps/inout/basic.py
@@ -447,6 +447,9 @@ class UrlHandler(FileHandler):
 
     @property
     def file(self):
+        """Downloads URL and return file pointer.
+        Checks if size is allowed before download.
+        """
         if self._file is not None:
             return self._file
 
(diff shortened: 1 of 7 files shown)


cehbrecht:
tests have troubles with gdal installation ...

cehbrecht:
Test CI was using `ubuntu-latest` with is now using gdal=3. As workaround I'm using `ubuntu-18.04`.

I open an issue for gdal to solve separately.

pywps/inout/basic.py
```diff
@@ -495,15 +498,16 @@ def _openurl(href, data=None):
...
         and units.
 
-        :return: maximum file size in bytes
+        :return: maximum file size in bytes and megabytes
         """
         ms = config.get_config_value('server', 'maxsingleinputsize')
-        return config.get_size_mb(ms) * 1024**2
+        mb_size = config.get_size_mb(ms)
+        byte_size = mb_size * 1024**2
+        return byte_size, mb_size
```
huard, line 510: I think it's unexpected that a function would return the same output with two different units. I suggest PyWPS enforces some kind of developer convention, that whenever file size is described by float or int, it's in units of bytes.

The string formatting can then be done with something like that: https://stackoverflow.com/questions/1094841/get-human-readable-version-of-file-size

Zeitsperre: If you're interested, I have a function or two I built based on some of those SO responses. https://github.com/Ouranosinc/miranda/blob/1808dd7de48a650527b0968326ff051d137077a1/miranda/storage.py#L234 and https://github.com/Ouranosinc/miranda/blob/1808dd7de48a650527b0968326ff051d137077a1/miranda/storage.py#L273. Feel free to copy and adapt.

cehbrecht: Right  👍  ... I look into this.

cehbrecht:
The cause for `FileSizeExceeded` was probably an unwanted download when using metalink with URL items. So I can remove the new parameter `maxsingleoutputsize` ... is is not necessary.

pywps/inout/basic.py
cehbrecht: Ok to use [humanize](https://pypi.org/project/humanize/)? We might want to do this more often ...

Zeitsperre: I had no idea this existed. I can definitely find a use for this library elsewhere. Great find!

pywps/inout/basic.py
```diff
@@ -495,15 +502,15 @@ def _openurl(href, data=None):
 
         return req
 
-    @staticmethod
```
Zeitsperre, line 498: This is no longer a staticmethod?

pywps/inout/basic.py
cehbrecht: 👍

pywps/inout/basic.py
cehbrecht: I can revert it to static ...

cehbrecht: done.

cehbrecht:
@huard @Zeitsperre merged. Need a patch release on monday. Please complain ... we can fix it with another patch release :)

PR #581, opened 2021-03-19
fix metalink size for url

cehbrecht:
# Overview

This PR adds a "size" checker in `URLHandler`. 

When the metalink document is created it is checking the "file" size of a metafile item. When the metafile is representing an external URL it is using the `URLHandler`.

Issue: when we were calculating the size of an "URL" it was first *downloading* the URL and the calculating the file size. We did not recognize ... files were small before ;) This is probably the root cause for issue #571. 

Now, the `URLHandler` and the `FileHandler` have their own methods to calculate "size". Avoiding unwanted downloads.

# Related Issue / Discussion

#571 #580 

# Additional Information

# Contribution Agreement

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,7 +4,7 @@ on: [ push, pull_request ]
 
 jobs:
   main:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
     strategy:
       matrix:
         python-version: [3.6, 3.7, 3.8, 3.9]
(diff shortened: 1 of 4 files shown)


huard (approved):
LGTM!

.github/workflows/main.yml
```diff
@@ -4,7 +4,7 @@ on: [ push, pull_request ]
 
 jobs:
   main:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
```
Zeitsperre, line 7: This is a good idea in general. I'm just curious, but why was this failing ?

Zeitsperre (approved):

.github/workflows/main.yml
Zeitsperre: Oh, It was GDAL. No surprise there. LGTM!

cehbrecht: yes, gdal again ;)

PR #582, opened 2021-03-20
Metalink: optimize size calculation

cehbrecht:
# Overview

This PR is an update of PR #581.

Changes:
* Using `_openurl` with requests stream to get content-length. Faster than `head` request.
* The `size` of a metafile can be set from outside to avoid calculation. Size might already be know and/or it takes long too calculate it when the metalink document has many files (>100).

# Related Issue / Discussion

#581 

# Additional Information

# Contribution Agreement


cehbrecht:
ping @huard @Zeitsperre

PR #584, opened 2021-03-21
using storage_copy_function

cehbrecht:
# Overview

This PR adds a new server option `storage_copy_function`. You can choose the `copy_function` in `FileStorage` when copying outputs to the destination folder: `copy`, `move` or `link`. Default=`copy` (as it was already).

Motivation: when having large output files we don't want to copy them ... preferred method is `link` making a hardlink using `os.link`. `move` would be an alternative using `shutil.move`. Fallback is always `copy` with `shutil.copy2`. 

# Related Issue / Discussion

# Additional Information

The `link` was also implemented in pywps 3.x:
https://github.com/geopython/pywps/blob/f18c0fdcb32cc354c856edde56fbdcb8db87394f/pywps/Wps/Execute/__init__.py#L1157-L1164

# Contribution Agreement

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -156,6 +156,15 @@ configuration file <https://docs.pycsw.org/en/latest/configuration.html>`_.
 :storagetype:
     The type of storage to use when storing status and results. Possible values are: ``file``, ``s3``. Defaults to ``file``.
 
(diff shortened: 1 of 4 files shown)


cehbrecht:
ping @Zeitsperre @huard

PR #589, opened 2021-03-28
show lineage also when process failed

### LINKED ISSUE #587: Return lineage info (inputs) also with error response?
# Description

In case of an error we would like to know the request. The wps has the `lineage` parameter to optionally return the inputs with a response document. Currently this happens only when the process was successful.

The [WPS 1.0.0 spec](https://www.ogc.org/standards/wps) says the following (p. 42):

> If lineage is ―true‖ (see Table 50), the Execute operation response shall include the DataInputs and OutputDefinitions elements. The server shall include in the execute response a complete copy of the DataInputs and OutputDefinition elements as received in the execute request. If lineage is "false" then these elements shall be omitted from the response.

# Environment

- operating system:
- Python version: 
- PyWPS version: 4.4.1
- source/distribution
 - [ ] git clone
 - [ ] Debian
 - [ ] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [ ] Apache/mod_wsgi
 - [ ] CGI
 - [ ] other (please specify): 

# Steps to Reproduce

# Additional Information

cehbrecht:
# Overview

With this PR the lineage info will also be written when the process has failed.

# Related Issue / Discussion

Fix #587 

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/response/execute.py
+++ b/pywps/response/execute.py
@@ -170,7 +170,10 @@ class ExecuteResponse(WPSResponse):
             data["status"] = self._process_paused()
         elif self.status == WPS_STATUS.SUCCEEDED:
             data["status"] = self._process_succeeded()
-
+            # Process outputs XML
+            data["outputs"] = [self.outputs[o].json for o in self.outputs]
+        # lineage: add optional lineage when process has finished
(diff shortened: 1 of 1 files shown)


Zeitsperre (approved):
Nice change.

PR #594, opened 2021-03-29
Make PyWPS Windows friendly

idanmiara:
# Overview

Resolved all the issues that prevented the tests to pass on Windows.

# Contribution Agreement


idanmiara:
Not sure why 3.6 failed and 3.7 got canceled...

idanmiara:
@cehbrecht Could you review please?

pywps/inout/basic.py
```diff
@@ -2,10 +2,13 @@
 # Copyright 2018 Open Source Geospatial Foundation and others    #
 # licensed under MIT, Please consult LICENSE.txt for details     #
 ##################################################################
+from typing import Union
 
 from pywps.translations import lower_case_dict
 from io import StringIO
 import os
+import pathlib
```
Zeitsperre, line 10:
```suggestion
```

Zeitsperre (commented):
Barring a few changes, looks good. I leave @cehbrecht to do the approval.

pywps/inout/basic.py
```diff
@@ -334,8 +337,8 @@ def mem(self):
     @property
     def url(self):
         """Return url to file."""
-        import pathlib
-        return pathlib.PurePosixPath(self.file).as_uri()
+        result = pathlib.PurePath(self.file).as_uri()
```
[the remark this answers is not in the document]
idanmiara: thanks

pywps/inout/basic.py
```diff
@@ -1033,3 +1036,22 @@ def get_url(self):
...
+def file_uri(path: Union[str, Path]) -> str:
+    path = Path(path)
+    path = path.as_uri()
+    return str(path)
+
+
+def uri_to_path(uri) -> str:
+    p = urlparse(uri)
+    path = p.path
+    if os.name == 'nt':
```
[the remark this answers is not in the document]
idanmiara: > This might not be enough to determine if the OS/Platform is actually Windows. I would leverage `platform.system()` for a few windows-like keys. See: https://stackoverflow.com/a/54837707/7322852

I looked at the post, I don't mind to follow your recommendation, but I couldn't find any system (windows-like or not) that have `os.name != 'nt' and platform.system() == 'windows'`

idanmiara:
> Barring a few changes, looks good. I leave @cehbrecht to do the approval.

Thanks for your review!

pywps/inout/basic.py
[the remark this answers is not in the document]
Zeitsperre: I bring it up here since depending on the subsystem that the script is running in (msys2, cygwin, WSL2, etc.), this check might not return what we intend it to. There's also a possibility that the `pathlib.Path` function handles the file path changes in a better way (see: https://docs.python.org/3/library/pathlib.html#pathlib.PureWindowsPath) e.g.:
```
from pathlib import Path, PureWindowsPath, PurePosixPath
p = Path("D/path/to/a/file")
p.as_posix()
Out[5]: 'D/path/to/a/file'
str(PureWindowsPath(p))
Out[9]: 'D\\path\\to\\a\\file'
PurePosixPath(PureWindowsPath(p))
Out[15]: PurePosixPath('D/path/to/a/file')
```

To be fair, I haven't done much in Windows. Just presenting some options.

Zeitsperre:
PyWPS seems to be failing for Python3.6. Looking into this now.

pywps/configuration.py
```diff
@@ -17,6 +17,8 @@
 
 __author__ = "Calin Ciociu"
 
+from pywps.inout.basic import file_uri
```
Zeitsperre, line 20: It appears that this is creating a circular import. You might need to find a different approach for loading this function (move to another file maybe)?

idanmiara: done, thanks!
Where did you find the error message?

Zeitsperre: You might still need to update some tests: 
```
 ImportError: cannot import name 'uri_to_path' from 'pywps.inout.basic' (/home/runner/work/pywps/pywps/pywps/inout/basic.py)
```

If you click on the checks below, you can see where each build is failing. Stack trace is there.

idanmiara: thanks! somehow I've missed that stack trace :)

pywps/inout/basic.py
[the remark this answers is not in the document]
idanmiara: Thanks for bringing this up.
Actually I've used it here, but the only thing I couldn't figure out is how to get a proper WindowsPath from a file uri:
```
> p = r'c:\abc.txt'
> wp = WindowsPath(p)
> wp
WindowsPath('c:/abc.txt')
> wp.as_uri()
'file:///c:/abc.txt'
> from urllib.parse import urlparse
> parse = urlparse(uri)
> parse.path
'/c:/abc.txt'
> Path(parse.path)
WindowsPath('/c:/abc.txt')
```
How can I get rid of this leading slash without checking actually trimming it if on windows ?

pywps/configuration.py
idanmiara: I don't understand why only in 3.6 I get this error now...
`AttributeError: module 'pywps' has no attribute 'configuration'`

idanmiara:
Finally it's ready!

Zeitsperre (approved):

cehbrecht:
I'm confused by the diff ... why is `tests/processes/metalinkprocess.py` shown as deleted and then added?

pywps/configuration.py
```diff
@@ -17,6 +17,8 @@
 
 __author__ = "Calin Ciociu"
 
+from pywps.uri import file_uri
```
cehbrecht, line 20: maybe we can have `pywps.util` with these helper functions?

Zeitsperre:
Oh! It's because the example process was originally in the docs and then was referred to within the `processes`. How strange! Good catch. This is a much better apporoach.

idanmiara:
> I'm confused by the diff ... why is `tests/processes/metalinkprocess.py` shown as deleted and then added?

As @Zeitsperre wrote, there was a symlink between the file at the docs to the source dir.
Symlinks are not compatible with Windows, so I've moved the file to the source dir and referenced it from the docs instead of a symlink.
I've tried to do it in one commit by git does not change the status the file from symlink to regular file, so I ended up splitting it this way.

cehbrecht (approved):
Only checked on linux ... it still works :)

cehbrecht:
@idanmiara @Zeitsperre I might use `pywps.util` instead of `pywps.uri` ... I leave the vote to you :) I will merge then.

pywps/configuration.py
idanmiara: good idea, I'll do that

pywps/tests.py
```diff
@@ -172,5 +174,6 @@ def assert_wps_version(response, version="1.0.0"):
                           '/ows:ServiceTypeVersion')
     found_version = elem[0].text
     assert version == found_version
-    with open("/tmp/out.xml", "wb") as out:
+    tmp = Path(tempfile.mkdtemp())
```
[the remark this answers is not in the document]
Zeitsperre: That would remove a lot of boilerplate, good suggestion. We can address that in another PR.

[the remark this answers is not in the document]
idanmiara: > That would remove a lot of boilerplate, good suggestion. We can address that in another PR.

agree 👍

pywps/configuration.py
idanmiara: > maybe we can have `pywps.util` with these helper functions?

done

idanmiara:
> @idanmiara @Zeitsperre I might use `pywps.util` instead of `pywps.uri` 

done.
Is the CI broken now ?

Zeitsperre:
@idanmiara It is but not because of you. The fix is quite simple if you want to add it. In `.github/workflows/main.yml`:```
sudo apt-get -y install gdal-bin libgdal-dev libnetcdf-dev libhdf5-dev
```
Should be changed to:```
sudo apt-get update && sudo apt-get -y install gdal-bin libgdal-dev libnetcdf-dev libhdf5-dev
```

I don't think either of us has access to the GithHub action. @cehbrecht ?

cehbrecht:
> @idanmiara It is but not because of you. The fix is quite simple if you want to add it. In `.github/workflows/main.yml`:
> 
> ```
> sudo apt-get -y install gdal-bin libgdal-dev libnetcdf-dev libhdf5-dev
> ```
> 
> Should be changed to:
> 
> ```
> sudo apt-get update && sudo apt-get -y install gdal-bin libgdal-dev libnetcdf-dev libhdf5-dev
> ```
> 
> I don't think either of us has access to the GithHub action. @cehbrecht ?

you can update it in this PR changing the workflow:
https://github.com/geopython/pywps/blob/pywps-4.4/.github/workflows/main.yml

Zeitsperre:
> > @idanmiara It is but not because of you. The fix is quite simple if you want to add it. In `.github/workflows/main.yml`:
> > ```
> > sudo apt-get -y install gdal-bin libgdal-dev libnetcdf-dev libhdf5-dev
> > ```
> > 
> > 
> > Should be changed to:
> > ```
> > sudo apt-get update && sudo apt-get -y install gdal-bin libgdal-dev libnetcdf-dev libhdf5-dev
> > ```
> > 
> > 
> > I don't think either of us has access to the GithHub action. @cehbrecht ?
> 
> you can update it in this PR changing the workflow:
> https://github.com/geopython/pywps/blob/pywps-4.4/.github/workflows/main.yml

Of course! I was looking at the `master` and got very confused as to why I couldn't find it. I forget that we don't base directly off master for this project.

idanmiara:
> > > @idanmiara It is but not because of you. The fix is quite simple if you want to add it. In `.github/workflows/main.yml`:
> > > ```
> > > sudo apt-get -y install gdal-bin libgdal-dev libnetcdf-dev libhdf5-dev
> > > ```
> > > 
> > > 
> > > Should be changed to:
> > > ```
> > > sudo apt-get update && sudo apt-get -y install gdal-bin libgdal-dev libnetcdf-dev libhdf5-dev
> > > ```
> > > 
> > > 
> > > I don't think either of us has access to the GithHub action. @cehbrecht ?
> > 
> > 
> > you can update it in this PR changing the workflow:
> > https://github.com/geopython/pywps/blob/pywps-4.4/.github/workflows/main.yml
> 
> Of course! I was looking at the `master` and got very confused as to why I couldn't find it. I forget that we don't base directly off master for this project.

done, now it works :)

PR #602, opened 2021-04-25
 Simplify the implementation of IOHandler 

gschwind:
# Overview

The current implementation use a tricks that manipulate the __mro__.
This trick do not have any obvious benefit and make the code much more
complicated than necessary. This patch simplifier the implementation by
replacing the tricky class change to a change in mamber's variable.

# Related Issue / Discussion

Following discussion in #598, Here is my proposal of refactoring the IOHandler. It come in two flavor, one with is a basic one and a second using weakref. I thought of another way using only class object without instantiating them like the previous implementation but I don't like too much the idea.

The code pass all the test excluding two. the two failed test a due to the fact that the test use private API to trick the test checking, thus IMO this just the test that is invalid.

Best regards.

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/inout/basic.py
+++ b/pywps/inout/basic.py
@@ -302,6 +302,12 @@ class IOHandler(object):
         return self._iohandler.base64
 
     @property
+    def size(self):
+        """Return object size in bytes.
+        """
+        return self._iohandler.size
+
+    @property
     def file(self):
         """Return a file name"""
         return self._iohandler.file
--- a/pywps/inout/basic.py
+++ b/pywps/inout/basic.py
@@ -356,18 +356,6 @@ class IOHandler(object):
     def prop(self):
         return self._iohandler.prop
 
-    @prop.setter
-    def prop(self, value):
-        LOGGER.warning("Deprecated use of IOHandler.prop, prefer self.value = self.value")
-        if value == "file":
-            self._iohandler = FileHandler(self._iohandler.file, self)
-        elif value == "data":
-            self._iohandler = DataHandler(self._iohandler.data, self)
-        elif value == "stream":
-            self._iohandler = StreamHandler(self._iohandler.stream, self)
-        elif value == "url":
-            self._iohandler = UrlHandler(self._iohandler.url, self)
-
 
 class FileHandler(NoneIOHandler):
     prop = 'file'
--- a/pywps/inout/basic.py
+++ b/pywps/inout/basic.py
@@ -33,6 +33,8 @@ from copy import deepcopy
 from io import BytesIO
 import humanize
 
+import weakref
+
 
 _SOURCE_TYPE = namedtuple('SOURCE_TYPE', 'MEMORY, FILE, STREAM, DATA, URL')
 SOURCE_TYPE = _SOURCE_TYPE(0, 1, 2, 3, 4)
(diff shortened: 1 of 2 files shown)


huard (approved):
Made some proposals to fix failing tests here: https://github.com/gschwind/PyWPS/pull/1

Otherwise this looks good and makes the code clearer I believe.

pywps/inout/basic.py
```diff
@@ -254,58 +295,100 @@ def clone(self):
...
+    def prop(self, value):
+        LOGGER.warning("Deprecated use of IOHandler.prop, prefer self.value = self.value")
+        if value == "file":
+            self._iohandler = FileHandler(self._iohandler.file, self)
+        elif value == "data":
+            self._iohandler = DataHandler(self._iohandler.data, self)
+        elif value == "stream":
+            self._iohandler = StreamHandler(self._iohandler.stream, self)
+        elif value == "url":
+            self._iohandler = UrlHandler(self._iohandler.url, self)
```
huard, line 370: Do we really need this ? Removing it does not create any test failure. I don't think this was supported previously.

huard:
I've tested this with emu + birdy and I'm getting 3 INTERNAL SERVER ERRORS. 

https://github.com/bird-house/birdy

pywps/inout/basic.py
gschwind: Hello huard,

At beginning I did not implement the prop setter but the test : test_inout.py:test_complex_input_stream need it.

Otherwise, I agree that we should remove it.

gschwind:
Hello huart,

I will have a look, best regards.

gschwind:
Hello huard,

I found the guilty :)

Thank you for your feedbacks

Best regards

cehbrecht:
@gschwind @huard merged . Thanks :)

huard:
Hi @gschwind 

I'm working on parallelism in PyWPS and I'm hitting an issue with the serialization of Process instances. The IO handlers include weakrefs that are not picklable, and so I'm not able to send processes to "workers".  Could we replace those weakref by something else that would be pickable?

gschwind:
Hello huard,

Why you do not use json serialization as it's done currently to store request in data base ?

Best regards.

huard:
I tried to add `__setstate__` and __getstate__` method using the `json` property and `from_json` method, but it broke pickling. Replacing the weakrefs by normal references solved this issue, but I wanted to know what were the downsides. I might be missing something obvious. I'll open a separate issue about PyWPS object serialization.

huard:
Ok thanks, see #658.

gschwind:
Hello,

After a quick check, it is for breaking circular references. for instance within the code IOHandler._iohandler is a refecence to FileHandler or DataHandler and those object as FileHandler._ref and DataHandler._ref that is linked to there owner IOHandler. If I do not use weakref here I have a reference loop where IOHandler own FileHandler or DataHandler, and those FileHandler or DataHandler own there parent IOHandler.

If you want as dirty fix, you can use prepare_serialization that remove the weak ref with None and fix_deserialize that rebuild the weak_ref. like:

```
def prepare_serialization(self):
   self._iohandler._ref = None

def fix_deserialize(self):
    self._iohandler._ref = weakref.ref(self)
```

Best regards

huard:
Many thanks for the explanation and suggestion.

PR #614, opened 2021-07-22
OGC API - Processes: BoundingBox, Complex and reference Input/Output + tests

idanmiara:
# Overview

This fix add support for BoundingBox, Complex inputs/outputs and extend the relevant docs + tests.

# Related Issue / Discussion

v4.5.0 release https://github.com/geopython/pywps/pull/613
main implementation https://github.com/geopython/pywps/pull/588
docs https://github.com/geopython/pywps/pull/612

# Contribution Agreement

### DIFF
--- a/docs/api_rest.rst
+++ b/docs/api_rest.rst
@@ -330,7 +330,76 @@ POST Execute Response:
     Hello Dude
 
 
-The example above only shows `LiteralInput`.
+Example for a reference input:
+
+.. code-block::
+
+    "raster": {
+        "type": "reference",
+        "href": "file:./path/to/data/data.tif"
+    }
+
+Example for a BoundingBox input:
+(bbox default axis order is yx (EPSG:4326), i.e. miny, minx, maxy, maxx)
+
+.. code-block::
+
+    "extent": {
+        "type": "bbox",
+        "bbox": [32, 34.7, 32.1, 34.8]
+    }
+
+
+Example for a ComplexInput input:
+(the data is a standard GeoJSON)
+
+.. code-block::
+
+    "cutline": {
+        "type": "complex",
+        "data": {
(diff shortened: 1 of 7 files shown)


idanmiara:
@cehbrecht I tried to support GeoJSON input also inline inside the request and also via a reference.
I stumbled upon the problem that if I used a reference then the data inside the ComplexInput is represented as a string (which we need to apply json.loads(s) to get the dict), whereas if I insert the data inline then the data insde the ComplexInput is a dict.

Trying to preserve maximum backwards compatibility, i.e not to break usage like:
https://github.com/geopython/pywps/blob/main/tests/test_complexdata_io.py#L117

I eventually inserted the json.loads logic inside the data_as_json function:

https://github.com/geopython/pywps/pull/614/commits/c054817225b842d1868574cba7b25e67745ebb62#diff-1504dae61cda8378b070b94031958e8f3a2c954f16959a33876f9411ec249409R323
https://github.com/geopython/pywps/pull/614/commits/8c188a44b7b2cd6596edb6e2c3b6e521b11f5767#diff-3dfc734e5dee6eba305a9939b9f1dd809679ae210baa3169b2a22814de8b061fR119

Do you have a more elegant way for doing so?

I think otherwise this PR is ready, and I'd be happy if it can make it into v4.5.

idanmiara:
Please review
@jachym @tomkralidis @huard @cehbrecht 
I'm not sure what's up with the @coveralls as I did add relevant tests (and docs :)

cehbrecht (approved):

cehbrecht:
@idanmiara Looks good to me :) We can look at improvements later on. I have another PR in the pipeline before making 4.5 release.

idanmiara:
Thanks @cehbrecht!

Issue #619, opened 2021-08-13
invalid mimetype


cehbrecht:
# Description

When sending a request with a compex output the pywps services may fail:

```
File "/usr/local/anaconda/envs/rook/lib/python3.7/site-packages/pywps/app/Process.py", line 478, in setup_outputs_from_wps_request
    'MimeType ' + mimetype + ' not valid')
TypeError: can only concatenate str (not "NoneType") to str
```

The default mimetype has changed from empty string `''` to `None` in pywps 4.5.0 which lets to this error:

https://github.com/geopython/pywps/blob/20147561c6985525c26b31ee9eac86192487b07e/pywps/app/WPSRequest.py#L525

# Environment

- operating system:
- Python version:
- PyWPS version: 4.5.0
- source/distribution
 - [ ] git clone
 - [ ] Debian
 - [ ] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [ ] Apache/mod_wsgi
 - [ ] CGI
 - [ ] other (please specify): 

# Steps to Reproduce

Run the [Emu](https://github.com/bird-house/emu) `inout` process with a post request (or using [birdy](https://github.com/bird-house/birdy)):

```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<wps:Execute service="WPS" version="1.0.0" xmlns:wps="http://www.opengis.net/wps/1.0.0" xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/wps/1.0.0
../wpsExecute_request.xsd">
	<ows:Identifier>inout</ows:Identifier>
	<wps:ResponseForm>
		<wps:ResponseDocument storeExecuteResponse="true" status="true">
			<wps:Output asReference="true">
				<ows:Identifier>output</ows:Identifier>
			</wps:Output>
		</wps:ResponseDocument>
	</wps:ResponseForm>
</wps:Execute>
```

# Additional Information

cehbrecht:
fixed by PR #620

PR #620, opened 2021-08-13
 fix None mimetype  

cehbrecht:
# Overview

This PR fixes issue #619. 

Changes:
* Updated `app/Process.py` to cope with `None` mimetype.
* Added test `test_assync_inout` to validate this patch.

# Related Issue / Discussion

#619 

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/app/Process.py
+++ b/pywps/app/Process.py
@@ -457,7 +457,7 @@ class Process(object):
             for wps_outpt in wps_request.outputs:
 
                 is_reference = wps_request.outputs[wps_outpt].get('asReference', 'false')
-                mimetype = wps_request.outputs[wps_outpt].get('mimetype', '')
+                mimetype = wps_request.outputs[wps_outpt].get('mimetype', None)
                 if is_reference.lower() == 'true':
                     # check if store is supported
                     if self.store_supported == 'false':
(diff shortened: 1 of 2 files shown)


tests/test_assync_inout.py
```diff
@@ -0,0 +1,69 @@
...
+            WPS.ResponseDocument(
+                WPS.Output(
+                    OWS.Identifier("text")
+                ),
+            ),
+        ),
+        version="1.0.0"
+    )
+    resp = client.post_xml(doc=request_doc)
+    print(resp.data)
```
Zeitsperre, line 64:
```suggestion
```

pywps/app/Process.py
```diff
@@ -471,9 +471,9 @@ def setup_outputs_from_wps_request(self, wps_request):
                 for outpt in self.outputs:
                     if outpt.identifier == wps_outpt:
                         outpt.as_reference = is_reference
-                        if isinstance(outpt, ComplexOutput) and mimetype != '':
+                        if isinstance(outpt, ComplexOutput) and mimetype is not None:
```
Zeitsperre, line 474: You could also simply check for mimetype, no? If it isn't `None` or `""`, it should evaluate to `True`.

Zeitsperre (approved):

pywps/app/Process.py
cehbrecht: Yes ... done.

huard (approved):

PR #629, opened 2021-09-28
resovle invalid Exception.msg unknown attribute

fmigneault:
# Overview

Exception object does not have `msg` attribute. The real exceptions are therefore not reported because they are masked away by raised `AttributeError`. 

Replace invalid `.msg` by `str()` operation to let it resolve the appropriate error cause string representation as implemented by any exception derived class.

# Contribution Agreement


cehbrecht (approved):

cehbrecht:
@fmigneault Thanks :) I will fix this also on 4.4.x branch.

PR #636, opened 2021-11-11
Create output directory before estimating available space on device

### LINKED ISSUE #635: Should `FileStorage.do_store` create the `outputpath` directory ?
# Description

I'm a bit confused on whether or not PyWPS is expected to create the `outputpath` directory if it does not exist. 

The documentation says *server path where to store output files.*

If the directory does not exist, PyPWS fails in this call [here](https://github.com/geopython/pywps/blob/ced1484299761a834cc1cc854f4d9e9c2856e8b2/pywps/inout/storage/file.py#L85)
but it feels like this failure could be avoided by computing the free space on the drive where the target resides, or the *closest* existing parent directory.  Later in the code, `os.makedirs` is called, suggesting an intent to create recursively the target path. 

So it's not clear to me exactly what the desired behavior is. 

# Environment

- operating system: Ubuntu
- Python version: 3.9
- PyWPS version: 4.4

huard:
# Overview

For file storage, the available space on device is computed before directory is created. This fails when the output path requires creating multi-level directories.

# Related Issue / Discussion
Fix #635 

# Additional Information

# Contribution Agreement


huard:
@cehbrecht Any idea about the GDAL error. Shouldn't it be updated to version 3 ?

tlvu:
@huard Does this create dir fix for `server.outputpath` also apply for `server.workdir` and `server.temp_path` in `wps.cfg`?  If not then we should also ensure these dir are created in not exist, see https://github.com/geopython/pywps/issues/635#issuecomment-956551246.

Otherwise, PR LGTM, thanks !

huard:
I don't think so. 
I haven't added a test yet to the PR. I could try to check that at the same time.

huard:
I added a test confirming that nested directory creation for stored objects work. 
Thanks @cehbrecht for the GDAL fix.

cehbrecht (approved):

cehbrecht:
@huard looks good :) Feel free to merge.

PR #639, opened 2021-11-23
Try to install gdal using the system version

huard:
# Overview
- Set default GDAL version to >3
- Get GDAL version number from system in CI workflow

# Related Issue / Discussion

# Additional Information

# Contribution Agreement


cehbrecht:
I was looking at a more radical approach #638 :)

huard:
Made unnecessary by #638

PR #647, opened 2022-03-13
Fix invalide use of encode on bytes object

gschwind:
# Overview

Invalide use of encode on bytes object

Within the pywps.inout.inputs.ComplexInput the json convertion function use encode where decode should be used. The patch fix this issue.

# Related Issue / Discussion

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/inout/inputs.py
+++ b/pywps/inout/inputs.py
@@ -243,7 +243,7 @@ class ComplexInput(basic.ComplexInput):
                 else:
                     # Otherwise we assume all other formats are unsafe and need to be enclosed in a CDATA tag.
                     if isinstance(self.data, bytes):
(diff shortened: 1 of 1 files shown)


cehbrecht:
@gschwind Thanks for the patch. Would you like to merge to the `main` branch?

gschwind:
Hello,

I rebased the patch over the main here b5d9dc34ecbacade7212b2d53f50f0f9b2bc74d8 . I do not know how do you want to proceed.

cehbrecht:
@gschwind I would prefer to merge it to the `main` branch ... so we don't need to make a 4.4.x release anymore. If this is ok for you and we cannot change the base branch in this PR anymore, then you could open a new PR with a reference to this one.

Issue #610, opened 2021-05-28
BoundingBox input types cannot be serialised to submit to Slurm cluster


agstephens:
# Description

When sending a Bounding Box input type to the WPS server, it fails with an error:

```
<!-- PyWPS 4.4.0 -->
<ows:ExceptionReport xmlns:ows="http://www.opengis.net/ows/1.1" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd" version="1.0.0">
  <ows:Exception exceptionCode="SchedulerNotAvailable" locator="" >
      <ows:ExceptionText>Could not submit job: Object of type Crs is not JSON serializable</ows:ExceptionText>
  </ows:Exception>
</ows:ExceptionReport>
```

This seems to happen when trying to submit jobs to a Slurm cluster - so serialisation is required before submitting to Slurm, and it cannot do that.

# Environment

- operating system: CentOS7
- Python version: Python 3.7.8
- PyWPS version: 4.4.0
- source/distribution
 - [ ] git clone
 - [ ] Debian
 - [x] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [ ] Apache/mod_wsgi
 - [ ] CGI
 - [x] other (please specify): nginx

# Steps to Reproduce

I am connecting from a GUI that posts the request, not sure how to reproduce.

# Additional Information

I was able to fix if I make these two hacks to `owslib` and `pywps`:

```
owslib/ows.py
=========

    238 class BoundingBox(object):
    239     """Initialize an OWS BoundingBox construct"""
    240     def __init__(self, elem, namespace=DEFAULT_OWS_NAMESPACE):
    241         self.minx = None
    242         self.miny = None
    243         self.maxx = None
    244         self.maxy = None
    245         self.crs = None
    246         self.dimensions = 2
    247         if elem is None:
    248             return
    249         val = elem.attrib.get('crs') or elem.attrib.get('{{{}}}crs'.format(namespace))
    250         if val:
    251             try:
ADDED LINE:    252                 val = 'urn:ogc:def:crs:OGC:2:84'
    253                 self.crs = crs.Crs(val)
    254             except (AttributeError, ValueError):
    255                 LOGGER.warning('Invalid CRS %r. Expected integer' % val)

pywps/app/WPSRequest.py
==================

[...]

PR #650, opened 2022-03-23
fix JSON encoding of CRS type (#610) 

geotom:
# Overview
Fixed the JSON encoding of the CRS type which crashed an execution request when the request was serialized

# Related Issue / Discussion
This should fix the problem described in #610 

# Additional Information
The type `owslib.crs.Crs` cannot be encoded by the default JSON encoding

# Contribution Agreement

### DIFF
--- a/pywps/app/WPSRequest.py
+++ b/pywps/app/WPSRequest.py
@@ -11,7 +11,6 @@ from werkzeug.exceptions import MethodNotAllowed
 from pywps import get_ElementMakerForVersion
 import base64
 import datetime
-from owslib.crs import Crs
 from pywps.app.basic import get_xpath_ns, parse_http_url
 from pywps.inout.inputs import input_from_json
 from pywps.exceptions import NoApplicableCode, OperationNotSupported, MissingParameterValue, VersionNegotiationFailed, \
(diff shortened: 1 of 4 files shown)


cehbrecht (changes requested):
Could you please add a test for the bbox?

geotom:
Yes, I will add a coverage test for this. Latest next week

geotom:
### New changes to my branch: [issue-610](https://github.com/geotom/pywps/tree/issue-610)

I have added tests and also fixed the issue at the source of the problem, which was:
When parsing BBox inputs, a owslib `BoundingBox `was created and its crs property was simply set as input['crs']. The problem is that the crs propery of the `BoundingBox` is of the type `owslib.crs.Crs`. This lead to a failing JSONEncoder. 
I removed the check in the JSONEncoder and made now sure that the crs is properly set as string when doing the input parsing.

Should I create a new PR for this @cehbrecht or is it enought?

cehbrecht (approved):

cehbrecht:
@geotom @tomkralidis should we use `getcodeurn()` instead of `str(bbox.crs)`?:
https://github.com/geopython/OWSLib/blob/2f55bee4514ab114ab43b5495d719c66fd5dce79/owslib/crs.py#L1848

geotom:
The difference between `getcodeurn()` instead of `str(bbox.crs)` is probably cosmetic as `str()` indirectly calls exactly this method. Not sure what you'd prefer.

tomkralidis:
+1 to use `getcodeurn()`.

cehbrecht (approved):

cehbrecht:
@geotom ... test is failing now. `None` value of `bbox.crs` is not handled.

geotom:
Should not fail anymore @cehbrecht

cehbrecht:
@geotom thanks :)

PR #651, opened 2022-03-24
Fix invalide use of encode on bytes object

gschwind:
Within the pywps.inout.inputs.ComplexInput the json convertion function
use encode where decode should be used. The patch fix this issue.

same as #647 but main branch as base

# Overview

# Related Issue / Discussion

# Additional Information

# Contribution Agreement

### DIFF
--- a/pywps/inout/inputs.py
+++ b/pywps/inout/inputs.py
@@ -247,7 +247,7 @@ class ComplexInput(basic.ComplexInput):
                 else:
                     # Otherwise we assume all other formats are unsafe and need to be enclosed in a CDATA tag.
                     if isinstance(self.data, bytes):
-                        out = self.data.encode(self.data_format.encoding or 'utf-8')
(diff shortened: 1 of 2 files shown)


cehbrecht (approved):

cehbrecht:
@gschwind Tests are all passing. Would you like to add a test to prove it is doing the right thing?

cehbrecht:
@gschwind Tests are all passing. Would you like to add a test to prove it is doing the right thing?

cehbrecht:
@gschwind Tests are all passing. Would you like to add a test to prove it is doing the right thing?

gschwind:
I will have a look to add test, best regards

gschwind:
I did a test but I'm not satisfied because the serialization-deserialition is not symmetric :)

and the test is not symmetric too.

gschwind:
I updated the test to only test the decoding, it should be fine now

PR #657, opened 2022-05-11
Make asynchronous process test run asynchronously

### LINKED ISSUE #656: test_async does not run in async mode
# Description

It looks like the test in `tests/test_async.py` does not run in async mode. Also, it only checks that the execute response has been accepted, not that the process actually completed. 

I'm able to start the test in async mode, but I'm not sure how to get the final response, as the statusLocation does not seem to get updated.  

# Environment

- operating system: linux
- Python version: 3.8.12
- PyWPS version: 4.5.2
- source/distribution
 - [x] git clone
 - [ ] Debian
 - [ ] PyPI
 - [ ] zip/tar.gz
 - [ ] other (please specify): 
- web server
 - [ ] Apache/mod_wsgi
 - [ ] CGI
 - [ ] other (please specify): 

# Steps to Reproduce

# Additional Information

huard:
# Overview
The test in test_async was not running asynchronously. This PR runs the test with processing mode set to `distributed`, and confirms the process completes. 

# Related Issue / Discussion
Fix #656

# Additional Information
I moved the Sleep process definition with the other test processes. 

I had trouble initially because I thought owslib would be able to read the status file. Unfortunately, owslib does not support status files on the local file system. 

# Contribution Agreement

### DIFF
--- a/tests/test_z_async.py
+++ b/tests/test_z_async.py
@@ -19,13 +19,12 @@ WPS, OWS = get_ElementMakerForVersion(VERSION)
 
 class ExecuteTest(unittest.TestCase):
     def setUp(self) -> None:
-        self.mode = configuration.CONFIG.get('processing', 'mode')
         configuration.CONFIG.set('processing', 'mode', 'distributed')
 
     def tearDown(self) -> None:
-        configuration.CONFIG.set('processing', 'mode', self.mode)
+        configuration.load_configuration()
 
-    def test_assync(self):
+    def test_async(self):
         client = client_for(Service(processes=[Sleep()]))
         wps = WPSExecution()
 
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -10,6 +10,7 @@
 * @jonas-eberle Jonas Eberle
 * @cehbrecht Carsten Ehbrecht
 * @idanmiara Idan Miara
+* @huard David Huard
 
 # Contributor to older versions of PyWPS (< 4.x)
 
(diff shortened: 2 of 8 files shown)


huard:
Note that for longer "sleep" periods, test_async interferes with test_dblog. I solved this by shortening the sleep period, but there should be a cleaner solution to isolate test_dblog.

huard:
The update_status calls from the async process do not seem to be captured by the logging database in the test suite. Ideas ?

gschwind:
Hello,

I did a quick code review, imo you should split dca06fb6518bc1e00bf7a3f54981496288c1005d

into:
Refactor test.create_sleep into test.Sleep
Change test_async to actually test asynchronous process.

and to split cbd1c29c1816f2ed21ee2bfe232e70a41328c448, with typo fix separated from the rest.

More over having a loop with proper status check instead of sleep(.5) maybe better in dca06fb6518bc1e00bf7a3f54981496288c1005d

tests/test_z_async.py
```diff
@@ -0,0 +1,59 @@
...
+                               inputs=[('seconds', '.01')],
+                               output=[('finished', None, None)],
+                               mode='async')
+
+        resp = client.post_xml(doc=doc)
+        wps.parseResponse(resp.xml)
+        assert_response_accepted(resp)
+
+        # Wait for process to complete. The test will fail otherwise, which confirms the process is asynchronous.
+        time.sleep(.5)
```
fmigneault, line 42: The check should be done explicitly to validate that the current status is still running with a new `checkStatus`, but is not yet `ProcessSucceeded`. 

If the process is somehow running in sync, this would not confirm that async was applied. It would just do a small wait before checking the status of an already completed process.

huard:
Replaced by #663 and #664

PR #664, opened 2022-09-19
Activate async test

huard:
# Overview
A unit test for asynchronous processes was not actually running asynchronously. The test now runs in async mode and will raise an error if it ever runs synchronously. 

# Related Issue / Discussion
In response to comments in #657

# Additional Information
Depends on #663 

# Contribution Agreement

### DIFF
--- a/tests/processes/__init__.py
+++ b/tests/processes/__init__.py
@@ -99,3 +99,38 @@ class BBox(Process):
         area = request.inputs['area'][0].data
         response.outputs['extent'].data = area
         return response
+
+
+class Sleep(Process):
+    """A long running process, just sleeping."""
+    def __init__(self):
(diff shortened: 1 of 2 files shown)


gschwind (changes requested):
Hello

Thank for the fix :)

The while loop is too simple, It need to handle a time out to avoid infinite waiting, and it also need check for Failed status.

Otherwise pytest is working, when test do not fail :)

Best regards.

huard:
Done

gschwind (approved):
Look good to me

Maybe you should capitalise commit messages :)

Best regards

cehbrecht (approved):

cehbrecht:
@huard thanks :)

PR #665, opened 2022-11-03
Fix tests/test_capabilities.py with correct check of request case

gschwind:
As far as I know, the OGC standard state that request parameters name such as 'service' or 'request' are not case sensitive, but the values of those are case sensitive [1].

[1] OGC 06-121r3 §11.5.1

# Overview

# Related Issue / Discussion

# Additional Information

# Contribution Agreement


huard (approved):

cehbrecht (approved):

cehbrecht:
@gschwind thanks :)

Issue #675, opened 2023-05-26
Improve env variable interpolation


nbrinckm:
# Description

While searching for a nice way to set configuration settings with env variables I searched a little bit in the documentation and struggled somehow with the current limitations.

While it is possible to read some (so far for the default configuration unspecified) values directly from the env variables, it doesn't allow us to set variables like the `server:url` from the env variables due to some default value that is set here https://github.com/geopython/pywps/blob/main/pywps/configuration.py#L76

However, the documentation also pointed at some point to the pycsw project that allows some basic env variable interpolation https://docs.pycsw.org/en/latest/configuration.html#using-environment-variables-in-configuration-files

My idea now is to copy this functionality & use its EnvInterpolation class for the config parser.

PR #676, opened 2023-05-26
675 extended env variable interpolation

nbrinckm:
# Overview

This is an example implementation on how the env interpolation could work.

# Related Issue / Discussion

See https://github.com/geopython/pywps/issues/675

# Additional Information

This is currently just the solutiuon as it is in the pycsw. But it still hope it could be helpful for the future.
I'm a little bit worried about the backward compatibility. But as the system doesn't expand values for that there are no env variables it should be okayish for use cases that used the $ syntax for their values in the past.

# Contribution Agreement

### DIFF
--- a/pywps/configuration.py
+++ b/pywps/configuration.py
@@ -27,6 +27,36 @@ LOGGER = logging.getLogger("PYWPS")
 wps_strict = True
 
 
+class EnvInterpolation(configparser.BasicInterpolation):
+    """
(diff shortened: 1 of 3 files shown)


cehbrecht (approved):

cehbrecht (approved):

cehbrecht:
@nbrinckm thanks 🙂

PR #684, opened 2023-09-19
Sanitize tests suite

gschwind:
# Overview

The patch suite sanitize tests by ensuring that all test run in his own temporary directory and use his own configuration. Moreover each test use their own sqlite database and do not use anymore the memory database. All test should run without side effect on other tests. The patch also ensure that no test file leak and cleanup created files during the tests (check /tmp after run tests before/after the patch suite). It is also setup pytest to run in importlib mode as advised by pytest web page and remove source file from the tests path to ensure that only pip installed files are tested.

The patch suite include #680 , #681 and #682 to work properly.

It's also fix some tests

# Related Issue / Discussion

# Additional Information

# Contribution Agreement

### DIFF
--- a/default-sample.cfg
+++ b/default-sample.cfg
@@ -1,12 +1,38 @@
+# Configuration file sample
+#
+# Comment line with leading '#' to fallback to the hardcoded default value
+#
+# The configuration is load in this order:
+# - First the hardcoded value are loaded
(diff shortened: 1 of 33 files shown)


cehbrecht (approved):

cehbrecht:
py37 will be skipped in future PR.

Issue #685, opened 2023-10-17
Input with optional `reference` in `UoM` raises `KeyError` if not a OGC unit


fmigneault:
Following definition is problematic:
https://github.com/geopython/pywps/blob/6293f3d5f69470b252863a838160f6959c83e5cf/pywps/inout/basic.py#L79-L80

This code makes it such that a definition such as the following will raise:
```python
from pywps.inout import LiteralInput

# explicit None reference
LiteralInput.from_json({"identifier": "test", "uoms": [{"uom": "seconds", "reference": None}]})
```
```pytb
Traceback (most recent call last):
  File "/home/francis/dev/conda/envs/weaver/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3526, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-13-50a089990430>", line 1, in <module>
    LiteralInput.from_json({"identifier": "test", "uoms": [{"uom": "seconds", "reference": None}]})
  File "/home/francis/dev/conda/envs/weaver/lib/python3.10/site-packages/pywps/inout/inputs.py", line 353, in from_json
    json_input_copy['uoms'] = [
  File "/home/francis/dev/conda/envs/weaver/lib/python3.10/site-packages/pywps/inout/inputs.py", line 354, in <listcomp>
    basic.UOM(uom['uom'], uom['reference'])
  File "/home/francis/dev/conda/envs/weaver/lib/python3.10/site-packages/pywps/inout/basic.py", line 73, in __init__
    self.reference = OGCUNIT[self.uom]
KeyError: 'seconds'
```
```python
# omitting the reference
LiteralInput.from_json({"identifier": "test", "uoms": [{"uom": "seconds"}]})
```
```pytb
Traceback (most recent call last):
  File "/home/francis/dev/conda/envs/weaver/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3526, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-14-17dec5faa665>", line 1, in <module>
    LiteralInput.from_json({"identifier": "test", "uoms": [{"uom": "seconds"}]})
  File "/home/francis/dev/conda/envs/weaver/lib/python3.10/site-packages/pywps/inout/inputs.py", line 353, in from_json
    json_input_copy['uoms'] = [
[...]

PR #686, opened 2023-10-28
fix invalid supported UOMs description

fmigneault:
# Overview

Fix process description literal UOMs reporting duplicates of the first default UOM instead of all supported UOMs.

For example, passing `uoms=['metre', 'feet']` to `LiteralInput` would cause it to report as follows in the WPS description.
The default UOM would be repeated for as many supported UOMs that were passed to the input.
```xml
<LiteralData>
  <ows:DataType ows:reference="http://www.w3.org/TR/xmlschema-2/#float">float</ows:DataType>
  <UOMs>
    <Default>
      <ows:UOM ows:reference="">m/s²</ows:UOM>
    </Default>
    <Supported>
      <ows:UOM ows:reference="">metre</ows:UOM>
      <ows:UOM ows:reference="">metre</ows:UOM>  <!-- error should be 'feet' -->
    </Supported>
  </UOMs>
</LiteralData>
```

At the same time, this fixes the optional `reference` attribute accordingly if it was not provided explicitly, or could not be mapped to a known unit of OGC URN. 

According to https://schemas.opengis.net/ows/1.1.0/owsDomainType.xsd
the `ows:UOM` which depends on `ows:DomainMetadataType`
shows `<attribute ref="ows:reference" use="optional" />`

But the previous check of `UOM` class caused it to check of `OGCUNIT` if not provided with an empty string.
The corresponding `if` in the template were also not consistent about reporting `reference`. 

# Related Issue / Discussion

Required by https://github.com/crim-ca/weaver/pull/541 
Required by https://github.com/crim-ca/weaver/issues/430
Fixes https://github.com/geopython/pywps/issues/685

# Contribution Agreement

- [x] I'd like to contribute bugfix to PyWPS. I confirm that my contributions to PyWPS will be compatible with the PyWPS license guidelines at the time of contribution.

### DIFF
--- a/pywps/inout/basic.py
+++ b/pywps/inout/basic.py
@@ -77,7 +77,7 @@ class UOM(object):
         self.reference = reference
 
         if self.reference is None:
-            self.reference = OGCUNIT[self.uom]
+            self.reference = OGCUNIT.get(self.uom, '')
 
     @property
     def json(self):
--- a/pywps/templates/1.0.0/describe/literal.xml
+++ b/pywps/templates/1.0.0/describe/literal.xml
@@ -2,11 +2,11 @@
                 {% if put.uom %}
                 <UOMs>
                     <Default>
-                        <ows:UOM {%if put.uom %}ows:reference="{{ put.uom.reference }}"{% endif %}>{{ put.uom.uom }}</ows:UOM>
+                        <ows:UOM {%if put.uom.reference %}ows:reference="{{ put.uom.reference }}"{% endif %}>{{ put.uom.uom }}</ows:UOM>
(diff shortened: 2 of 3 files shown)


cehbrecht (approved):

cehbrecht:
@fmigneault with PR #687 the tests should work again. If you think this PR is ready I can merge it.

fmigneault:
@cehbrecht 
Can you consider this fix into 4.6.0 release?

fmigneault:
On my side, the tests were running, but you can combine them with the other PR to validate they work together. 
Thanks!

PR #698, opened 2025-12-15
Mark tests that require network or geotiff dependency.

sebastic:
# Overview

Add pytest markers to enable skipping certain tests like those that require network or certain dependencies.

# Additional Information

The Debian package build environment doesn't have network, tests requiring network need to be skipped.

The geotiff module is also not available.

# Contribution Agreement

### DIFF
--- a/tests/validator/test_complexvalidators.py
+++ b/tests/validator/test_complexvalidators.py
@@ -69,6 +69,7 @@ class ValidateTest(TestBase):
 
         return fake_input
 
+    @pytest.mark.online
     def test_gml_validator(self):
         """Test GML validator
         """
@@ -79,6 +80,7 @@ class ValidateTest(TestBase):
         # self.assertTrue(validategml(gml_input, MODE.VERYSTRICT), 'VERYSTRICT validation')
         gml_input.stream.close()
 
+    @pytest.mark.online
     @pytest.mark.xfail(reason="gml verystrict validation fails")
     def test_gml_validator_verystrict(self):
         """Test GML validator
@@ -117,6 +119,7 @@ class ValidateTest(TestBase):
         self.assertTrue(validateshapefile(shapefile_input, MODE.STRICT), 'STRICT validation')
         shapefile_input.stream.close()
 
+    @pytest.mark.geotiff
     def test_geotiff_validator(self):
         """Test GeoTIFF validator
         """
@@ -141,6 +144,7 @@ class ValidateTest(TestBase):
         else:
             self.assertFalse(validatenetcdf(netcdf_input, MODE.STRICT), 'STRICT validation')
 
+    @pytest.mark.online
     @pytest.mark.xfail(reason="test.opendap.org is offline")
     def test_dods_validator(self):
         opendap_input = ComplexInput('dods', 'opendap test', [FORMATS.DODS,])
--- a/pytest.ini
+++ b/pytest.ini
@@ -2,3 +2,6 @@
 addopts = --import-mode=importlib
 pythonpath = tests
 testpaths = tests
+markers =
+    online: marks tests requiring network
+    geotiff: marks tests requiring geotiff module
--- a/tests/test_execute.py
+++ b/tests/test_execute.py
@@ -235,6 +235,7 @@ def get_output(doc):
 class ExecuteTest(TestBase):
     """Test for Exeucte request KVP request"""
 
+    @pytest.mark.online
     @pytest.mark.xfail(reason="test.opendap.org is offline")
     def test_dods(self):
         if not WITH_NC4:
--- a/tests/test_inout.py
+++ b/tests/test_inout.py
@@ -15,6 +15,7 @@ import tempfile
 import json
 from pywps import inout
 import base64
+import pytest
 
 from pywps import Format, FORMATS
 from pywps.app.Common import Metadata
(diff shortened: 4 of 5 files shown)


tests/validator/test_complexvalidators.py
```diff
@@ -117,6 +119,7 @@ def test_shapefile_validator(self):
         self.assertTrue(validateshapefile(shapefile_input, MODE.STRICT), 'STRICT validation')
         shapefile_input.stream.close()
 
+    @pytest.mark.geotiff
```
Zeitsperre, line 122: Why not simply add the `geotiff` library to the test requirements?

sebastic: > Why not simply add the `geotiff` library to the test requirements?

Because the library is not packaged, as mentioned in the OP:
> The geotiff module is also not available.

Zeitsperre: I'm not sure what you mean.
PyPI: https://pypi.org/project/geotiff/ 
Conda: https://github.com/conda-forge/python-geotiff-feedstock

`geotiff` is a core dependency as well: https://github.com/geopython/pywps/blob/04f734b9721e764dc244f1de0cc5738f5743e41c/requirements.txt#L4

sebastic: > I'm not sure what you mean.

There is no Debian package for the geotiff Python library:
```
$ apt-file search dist-packages/geotiff ; echo $?
1                                         
```
See also: https://lists.debian.org/debian-gis/2025/12/msg00011.html

> `geotiff` is a core dependency as well: 

It is not imported unconditionally:

 https://github.com/geopython/pywps/blob/main/pywps/validator/complexvalidator.py#L332-L338

Zeitsperre: Ah, this is specifically a Debian issue! My bad.

If we want to have this level of change, I would suggest that we remove `geotiff` from the core dependencies and add it to an `extras` for tracking optional dependencies.

Zeitsperre: Alternatively, what prevents us from packaging `geotiff` on Debian? Lack of time/effort/bandwidth?

sebastic: > If we want to have this level of change, I would suggest that we remove `geotiff` from the core dependencies and add it to an `extras` for tracking optional dependencies.

Yes, please. Treat `fiona`, `geotiff`, and `netCDF4` in the same way.

> Alternatively, what prevents us from packaging `geotiff` on Debian?

Who is "us"?

Someone will need to do the work, which in this case should be fairly simple as mentioned on debian-gis@.
All dependencies from [`requirements.txt`](https://github.com/KipCrossing/geotiff/blob/main/requirements.txt) are packaged in Debian, but note that the [zarr package](https://tracker.debian.org/pkg/zarr) is newer than claimed to be supported (3.1.5):
```
zarr>=2.12.0,<2.13
```

> Lack of time/effort/bandwidth?

To quote myself from the mailinglist post I linked:
> I can't find the motivation to package an optional dependency and commit to its maintenance for a package I don't actually use myself.
> 
> We really need to get more geopython people involved in packaging, since Johan v/d Wauw stopped contributing all of those packages are on life support not getting the attention they need.

cehbrecht: @sebastic @Zeitsperre @ldesousa can we limit support to conda-forge? I'm not sure if anyone is using direct installation on debian based dists.

fmigneault: If core dependencies are adjusted to remove format-specific libraries, that will help reduce the package, which is great since not every deployment employs all combinations of data formats for their processing. 

However, if this change is done, it should be marked as a higher release than a simple patch, since deployments relying on them could break.

Another thing to clarify in the release notes would be that these libraries are actually required only if using ">=STRICT" validation mode (they are imported under them in https://github.com/geopython/pywps/blob/main/pywps/validator/complexvalidator.py), so the formats could still be employed in I/O definitions with lower validation if the libraries are not installed.

sebastic: > @sebastic @Zeitsperre @ldesousa can we limit support to conda-forge? I'm not sure if anyone is using direct installation on debian based dists.

OSGeo-Live is using the Debian package currently:

 https://github.com/OSGeo/OSGeoLive/blob/master/bin/install_pywps.sh#L53

That can of course be changed, I can then stop spending time on the pywps package in Debian.

CC: @kalxas

Zeitsperre: `zarr` v3 support has been notoriously difficult to adopt due to some very breaking changes. Most projects I work on are pinned below that version. `geotiff` would need to address this issue before anything.

> Who is "us"?

I've worked on `geotiff` as well and could try my hand at packaging it for Debian (for personal learning purposes). There are other packages (non-Python) I work on that I'd love to see in Debian/Ubuntu.

> Yes, please. Treat fiona, geotiff, and netCDF4 in the same way.

I can justify putting some work towards this for this week. Fun end of year project, maybe. @sebastic Would you be fine with me cherry-picking your commit here?

sebastic: > > Who is "us"?
> 
> I've worked on `geotiff` as well and could try my hand at packaging it for Debian (for personal learning purposes). There are other packages (non-Python) I work on that I'd love to see in Debian/Ubuntu.

You're very welcome to join the Debian GIS team:

 https://debian-gis-team.pages.debian.net/policy/index.html#membership

> > Yes, please. Treat fiona, geotiff, and netCDF4 in the same way.
> 
> I can justify putting some work towards this for this week. Fun end of year project, maybe. @sebastic Would you be fine with me cherry-picking your commit here?

Sure, go ahead.

PR #699, opened 2025-12-19
Split extra libraries and add more granular control over tests

Zeitsperre:
# Overview

This Pull Request splits out the libraries that support some complex validators (geotiff, netcdf4, GDAL/OGR, zarr) so that basic data types are supported by default.

It also modifies some of the test controls so that `pytest` markers can be used to explicitly select/deselect tests that rely on these extra libraries.

Finally, coverage is now done across all builds in parallel, using `tox-gh` and the `coverallsapp/github-action` to remove the `coveralls-python` library.

# Related Issue / Discussion

Changes here build on discussions found in #698

Closes #698 

# Additional Information

There are many more changes I'd like to implement, but I'll limit things here for now until discussion is had in #700.

# Contribution Agreement


Zeitsperre:
@cehbrecht I'm not certain why the coverage status is taking such a huge hit. I did remove a few useless imports, so it could be that the original values were too high?

fmigneault:
@Zeitsperre 
Potentially because the package were installed by default, their tests were running and checking some code paths (like their input format validation for example). Now, they would be skipped if not installed, therefore not "covered".

Zeitsperre:
> Now, they would be skipped if not installed, therefore not "covered".

I have split the builds with `tox` to report collective coverages based on the sum from lines covered by default dependencies and the extra dependencies. The logic for both configurations (with and without extras) is being hit 

There were a few tests that I disabled/skipped because there were no assertions made with the results: smoke tests, essentially. It's likely from those changes that the coverage is being impacted. In any case, I think the coverage score is fine.

Zeitsperre:
@fmigneault I figured out why the coverage was much higher: Previously, the values were reflective of both the source code (`pywps`) and the tests. The fact that the values are now restricted to the library, we lost around 4% of average coverage.

The value currently is also impacted by the removal of a handful of smoke tests, like I mentioned, but on the order of ~1%. Getting that back would require request mocking and other tricks. Are you familiar with those approaches?

cehbrecht (approved):

cehbrecht:
@Zeitsperre thanks for working on this :) This PR includes the merged #701? Does this PR needs to be re-based? I suppose we can ignore the coverage decrease ...

Zeitsperre:
@cehbrecht It does include #701, yes. No need to rebase as it was already up-to-date with the main branch.

I imagine the process here would be to merge this and then open another PR to perform the update actions (changelog, bumpversion)? The changes significant, but it should have very little impact on end-users.

tests/test_async.py
```diff
@@ -27,7 +26,7 @@ def setUp(self) -> None:
         # Running processes using the MultiProcessing scheduler and a file-based database
         configuration.CONFIG.set('processing', 'mode', 'distributed')
 
-    @pytest.mark.xfail(reason="async fails")
```
sebastic, line 30: The test doesn't only fail on macOS, it also fails for the Debian package build (with a 500 Internal Server Error).

Zeitsperre: I see. Is this the only failure? Are things better otherwise?

sebastic: Yes, it's the only failure.

We use `-m "not online and not requires_geotiff"` to skip tests that would fail in the package build environment.

I can't tell if things are better otherwise.

Zeitsperre: OK. I'll adjust this test and push a patch fix early next week. Hopefully this will solve the issues on your end.

PR #702, opened 2026-02-16
update LAS/LAZ point cloud formats to registered IANA media-types

fmigneault:
# Overview

The LAS/LAZ point cloud formats used to employ the generic `application/octet-stream`.
The preferred and registered IANA media-types should be used.

This avoids applications using `application/octet-stream` as "arbitrary data" to associated it to point cloud formats, which might not apply and break downstream processes that would detect `.las`/`.laz` extensions. 

# Related Issue / Discussion

n/a

# Contribution Agreement


cehbrecht (approved):
I'm ok with it  ... when tests pass :)

fmigneault:
@cehbrecht fixed with `main` sync

